PR #35, opened 2023-12-16
Allow running KLayout DRC script without GUI

Syndace:
This is a very small change to check whether `main_window` is valid before trying to get the current view. This allows the script to be run in batch mode (`klayout -zz -r sg13g2.lydrc <design>.gds`) without GUI. This is useful to run the DRC check in a script (for example as part of OpenROAD-flow-scripts) or on a remote server.


Syndace:
Closing in favour of https://github.com/IHP-GmbH/IHP-Open-PDK/pull/36

PR #104, opened 2024-05-06
Liberty view for IO cells update. 

KrzysztofHerman:
The updated view contains definition of all IO cells provided in the library hence it is feasible to use it in ORFS to design a pad ring.

PR #105, opened 2024-05-06
sg13g2_io: verilog: Add more cells

dnltz:
All IO cells, even simple Filler cells, have to be defined for chip-level simulations. Add more dummy cells for Corner and Filler cells.


dnltz:
@KrzysztofHerman figured out even more cells are required for chip-level simulations.

KrzysztofHerman:
Thank you for contribution, it will be added in the PR #104

PR #113, opened 2024-05-16
Proposal for making DRC independent from DBU setting

klayoutmatthias:
I received a support request for a user who used a DBU not equal to 1nm (2.5nm).

The DRC complains, so that is clearly a user error :)

However, I think it is possible to rewrite the DRC to metric units, so it is independent from DBU. This PR is a *proposal*. I have not fully tested it myself yet. But I tried on the user's sample and it appears to give reasonable results even with the wrong DBU.

Because I saved it in KLayout's editor, the file is formatted differently from the original. You can see the differences if turning on "ignore whitespace".

I am aware that DRC is generated, so applying the idea may not be feasible or a different approach is required. Please feel free to reject this PR in that case.

Best regards,

Matthias

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/drc/sg13g2.lydrc
+++ b/ihp-sg13g2/libs.tech/klayout/tech/drc/sg13g2.lydrc
@@ -1,29 +1,15 @@
-<?xml version='1.0' encoding='utf-8'?>
-<!--
- Copyright 2023 IHP PDK Authors
-
- Licensed under the Apache License, Version 2.0 (the "License");
- you may not use this file except in compliance with the License.
- You may obtain a copy of the License at
-
-     https://www.apache.org/licenses/LICENSE-2.0
-
- Unless required by applicable law or agreed to in writing, software
- distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


akrinke:
Thank you for this suggestion! We can apply your changes when generating the script. However, would IHP accept a layout with a DBU ≠ 1000?

sergeiandreyev:
let me check internally, but I guess such GDS will be rejected..

klayoutmatthias:
I agree, such a GDS should be rejected on Tapeout. 

In this particular case, the verification was for standard cells, not on full Tapeout. I understood that was a intermediate state. Finally of course, also standard cells should be also compliant with the DBU.

The topic is not urgent and it is okay for me to reject this PR.

I just would like to point one particular case (because I suffered from something like this): if - hypothetically - you would require a non-nm DBU like 0.25nm, then IMHO it is an huge advantage to write a DRC deck in metric units instead of DBU. Otherwise it is confusing to see `layer.width(800)` when you actually mean `layer.width(200.nm)`.

Best regards,

Matthias

akrinke:
After the merge of #147, this pull request can be closed. The new DRC script uses micrometers for all lengths.

PR #117, opened 2024-05-18
libs.teck: klayout: pycell: Remove filler from Sealring

dnltz:
Metal fill in the sealring was not DRC conform and should be handled by other tools.


dnltz:
Generates sealrings like shown here

image

sergeiandreyev:
Thank you, @dnltz! For this very needed contribution!

PR #120, opened 2024-05-25
Fixing paths and making PDK easier to install in Qucs-S

andresmmera:
The aim of this pull request is to facilitate the setup of the PDK in Qucs-S. 
I installed the PDK and tested a few things, but it took me a while to get it up and running, so I think the following commits will be useful:

***
`Commit #1`: **Copy sample schematics to Qucs-S home directory**
 
This adds a copy of the sample schematics to the Qucs home directory. By default, that’s where the user has his projects
and I find it very convenient to have the examples there. In order to make these files recognisable by Qucs-S, there must be located inside a folder with a name ending in “_prj”. I have given it the name “IHP-Open-PDK-SG13G2-Examples_prj” as it is something meaningful to the user.

image
image

***

`Commit #2`: **Inform the user to complete the PDK installation on the Qucs-S side**
 
This commit adds  information for the user on how to configure Qucs-S to work with the SG13G2 PDK. Qucs-S needs to know
where the libraries are located, so the user should should specify this manually. A new file inside the examples folder has been added with instructions on how to do it.

image
image

***

`Commit #3`: **Refer the library paths to the linux user’s home folder**
 
This modifies the schematics and the libraries to refer them to the user home. Prior to this, the schematics and the library files
included relative jumps. This means that the simulation can't find the symbols and models in the default configuration.

image

***

`Commit #4`: **Fix dc_diode_op.sch and dc_hbt_13g2.sch examples**
 
The commit is a fix for two schematics that have problems with the rectangular plot variables

### DIFF
--- a/ihp-sg13g2/libs.tech/qucs/examples/README.txt
+++ b/ihp-sg13g2/libs.tech/qucs/examples/README.txt
@@ -1,0 +1,9 @@
+####################################################
+				IMPORTANT
+####################################################
+
+In order to use the SG13G3 PDK sample schematics, you must add the SG13G3 PDK library to the Qucs-S subcircuit search path:
+
+1) Go to File ->Application Settings ...
+2) Click on the "Locations" tab
+3) On the bottom, click on the "Add Path With Subfolders" button and select the SG13G3 PDK library folder (/home/<username>/.qucs/user_lib)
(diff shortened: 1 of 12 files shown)


KrzysztofHerman:
Hi @andresmmera thank you for your contribution. Since we have made a lot of progress in the PDK we could not directly merge your PR however we "recycled" a vast majority in PR #162

PR #121, opened 2024-05-27
libs.ref: gds: Fix short between IOPadVdd and IOPadIOVdd

dnltz:
Placing both cells into an IO cells creates a short between both cells. This GDS contains a fix to not short them anymore.

`$ md5sum sg13g2_io.gds
21636651c59f5a4589508acb047e58a8  sg13g2_io.gds
`


dnltz:
@stafverhaegen-chipflow @KrzysztofHerman this is the gds file from last week.

sergeiandreyev:
just to confirm, this was the IO cells layout/GDS that was used for current tape-out?

dnltz:
> just to confirm, was this the IO cells layout/GDS that was used for current tape-out?

@sergeiandreyev Yes

PR #140, opened 2024-06-12
Update Filler Checks

dnltz:
The current DRCs don't report fill correctly and generate violations for non-chip areas. This PR also includes @KrzysztofHerman script to report the density.


akrinke:
The changes to the DRC scripts look goot, but I'm biased since I wrote them 😄
They will also be included in the next version of the DRC script.

dnltz:
> The changes to the DRC scripts look goot, but I'm biased since I wrote them 😄 They will also be included in the next version of the DRC script.

ah, I was not able to find your github handle! I mentioned you in the commit message ;)

PR #151, opened 2024-06-26
libs.tech: Klayout: tech: Add layer map

dnltz:
OpenROAD requires the sg13g2.map file to correctly place each layer. Fix this here to not overwrite manually changes in OpenROAD-Flow-Scripts.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/sg13g2.lyt
+++ b/ihp-sg13g2/libs.tech/klayout/tech/sg13g2.lyt
@@ -84,7 +84,7 @@
    <macro-resolution-mode>default</macro-resolution-mode>
    <separate-groups>false</separate-groups>
    <map-file>sg13g2.map</map-file>
-   <lef-files/>
+   <lef-files></lef-files>
   </lefdef>
   <mebes>
    <invert>false</invert>
--- a/ihp-sg13g2/libs.tech/klayout/tech/sg13g2.lyt
+++ b/ihp-sg13g2/libs.tech/klayout/tech/sg13g2.lyt
@@ -83,8 +83,8 @@
    <read-lef-with-def>true</read-lef-with-def>
    <macro-resolution-mode>default</macro-resolution-mode>
    <separate-groups>false</separate-groups>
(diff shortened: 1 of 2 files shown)


sergeiandreyev:
Hi @dnltz, is it possible to just put the `.map` file to this repo? in `libs.tech/klayout/tech` along with `.lyt` and `.lyp` files
I'm not sure that's a good idea to have this long relative path in KLayout tech file, referencing another repository / folder

dnltz:
> Hi @dnltz, is it possible to just put the `.map` file to this repo? in `libs.tech/klayout/tech` along with `.lyt` and `.lyp` files I'm not sure that's a good idea to have this long relative path in KLayout tech file, referencing another repository / folder

The path is actually used by OpenROAD to load the .map file. However, seems like no other PDK is using a .map file ... so maybe this can a workaround until a better solution for the .map file was found?

gadfort:
@dnltz we use the layer map file as well (although we are not fully enabled for this PDK yet). https://github.com/siliconcompiler/lambdapdk and https://github.com/siliconcompiler/siliconcompiler

sergeiandreyev:
there was a suggestion by Matthias Koefferlein (KLayout developer) last year to use the map file instead of modifying the technology `.lyt` file because the approach inside klayout is not efficient
the question is where to store / have this file

sergeiandreyev:
I prefer to have it as part of IHP OpenPDK repository, because it's PDK specific data, we set the correct licenses, we're aware of changes and can confirm or reject them, etc.

dnltz:
> I prefer to have it as part of IHP OpenPDK repository, because it's PDK specific data, we set the correct licenses, we're aware of changes and can confirm or reject them, etc.

alright, let me try to get this done without this OpenROAD specific path.

dnltz:
@sergeiandreyev with some search and replace magic in ORFS, the following `.lyt` file works.

@gadfort Are you okay with an empty `lef-files` tag? Your change broke ORFS ...

gadfort:
@dnltz yeah, an empty tag like that works for me.

PR #162, opened 2024-07-16
NGSPICE model referencing based on .spiceinit

KrzysztofHerman:
Major change in the operation of the ngspice and PDK models.
A .spiceinit was added under libs.tech/ngspice/. Since the pdk directory is referenced there it allows to stream the netlist without user specific location of the .lib files ie. .lib cornerMOSlv.lib mos_tt
The second change refers to the central location of the psp103_nqs.osdi model which is now situated in libs.tech/ngspice/openvaf. There is no need to use pre_osdi psp103_nqs.osdi in the spice netlist anymore.

The above changes affected xschem and qucs-s examples, which were also updated. In the case of qucs-s also the examples are copied to the user's .qucs directory and appear in the GUI in the project left pane. Some solutions were adopted from PR https://github.com/IHP-GmbH/IHP-Open-PDK/pull/120.


hpretl:
@KrzysztofHerman Please adapt the path construction:

Instead of `$PDK_ROOT/ihp-sg13g2/libs.tech/ngspice/models` use `$PDK_ROOT/$PDK/libs.tech/ngspice/models`!

hpretl:
@KrzysztofHerman BTW, I think this should be done generally throughout the PDK. So far I only see it also in the Qucs-S installation script.

ihp-sg13g2/libs.tech/ngspice/.spiceinit
sergeiandreyev, line 1: you have two times `+ option list`

PR #175, opened 2024-08-08
libs.ref: Python tool to generate export.yml

dnltz:
ORFS has an update script with a static list of files to update [1]. Everytime new files will be created, this script has to be updated.

Instead, let the PDK communicate to external tools which files exist, where they are located in the PDK and where they should be stored in the 3rd-party tool. This allows to automate updates in external tools automatically by just updating the PDK.

So, to push updated files to external tools, simply run `generate.py` and commit the new `export.yml` files. In case new files were created, update the list of source files in `generate.py` and trigger an update of `export.yml`.

1:  https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/platforms/ihp-sg13g2/sg13g2_update.py


sergeiandreyev:
Hi @dnltz, thank you for contribution!
A couple of comments:
- how about storing these files in `libs.tech/openroad`? this way it could be easier to understand what are these for, and we'll not impact the standard structure
- is it possible to merge all separate descriptions of PVT-specific Liberty files into one?
```
    {
        "file": "lib/sg13g2_stdcell_fast_1p32V_m40C.lib",
        "orfs": {"destination": "lib"}
    },
    {
        "file": "lib/sg13g2_stdcell_fast_1p65V_m40C.lib",
        "orfs": {"destination": "lib"}
    },
...
```

dnltz:
@sergeiandreyev also rebased to latest version of dev branch. I was using an older rev.

PR #189, opened 2024-08-29
Xycetestbenches - structure check

KrzysztofHerman:

- [ ] Tests pass


ihp-sg13g2/libs.tech/ngspice/.spiceinit
sergeiandreyev, line 1: why you have these changes in `.spiceinit`?
what we found out with Prabhat last week is that this has an impact on IO cells simulations

sergeiandreyev:
on the structure - I propose to put the `.py` and `.sh` scripts in `/scripts` folder

PR #193, opened 2024-09-03
Xyce adms lib

dwarning:
This is the first contribution to activate IHP-Open-PDK for Xyce.
psp103.va was modified to fulfil Xyce specific requirements and the special kind of sg13g2 AD, AS, PD, PS calculation.
As a starting point the moslv models and few Xyce simulation examples are provided.
The linux binary Xyce_Plugin_PSP103_VA.so should be removed in the final stage.
moshv and statistically examples should follow.


dwarning:
Thanks,
I think you should also remove PR #194 even the base is correct. Then I will come with an new PR for both.

PR #202, opened 2024-09-17
Improve Windows compatibility of PyCell code

akrinke:
This pull request removes the dependency on psutil, which is difficult to install under Windows. It would have to be compiled with the same compiler as KLayout.


ThomasZecha:
The current solution uses psutil, which supports a wide range of platforms. This PR approach to bypass the psutil installation-problem under Windows should not affect other platforms and shall be build as a specialization for Windows only. Additionally the use of the deprecated wmic-tool should be replaced by using the power-shell functionality.

PR #206, opened 2024-09-22
PyCell Klayout integration

ThomasZecha:
-fixed pypreprocessor issues

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
@@ -26,8 +26,6 @@ from cni.dlo import PCellWrapper
 # Creates the SG13_dev technology
 from .sg13_tech import *
 
-import pypreprocessor.pypreprocessor as preProcessor
-
 import pya
 
 import os
@@ -36,6 +34,7 @@ import sys
 import inspect
 import re
 import importlib
+import importlib.util
 import pathlib
 import tempfile
 
@@ -158,6 +157,9 @@ class PyCellLib(pya.Library):
                 isFirst = False
             print(f'Current process chain: {processChain}')
 
+        module = importlib.import_module(f"{__name__}.ihp.pypreprocessor")
+        preProcessor = getattr(module, "preprocessor")
+
         definesSetToPrint = []
 
         for moduleName in moduleNames:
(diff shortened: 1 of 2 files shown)


akrinke:
Sorry for the trouble that was caused by my pull request to improve Windows support. I see you have restored the symlink to `pyprocessor.py`. I have removed it, because I couldn't get it to work on Windows. Symlink support on Windows requires the activation of "Developer Mode" (see [here](https://stackoverflow.com/questions/5917249/git-symbolic-links-in-windows) and [here](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/#oTepts0M8148xRoj.97)) which I couldn't do on my work machine.

I had tested my solution and it seemed to work on Windows and Linux. What problems did you have with it? Do you think there is a way to remove the symlink?

ThomasZecha:
The link was needed to 'force' the right namespace when the pypreprocessor loads a PyCell which uses the conditional compilation feature (#ifdef). May you find a better solution without using this link?

akrinke:
Thanks for the explanation, I will have a look.

akrinke:
As a test, I added the following code to `inductors_code.py`:

```python
#ifdef klayout
    print("inside #ifdef")
#endif
```

This worked with the old code. Could you provide a short example that triggers the problem with the namespace?

ThomasZecha:
What do you mean with 'old code'?

akrinke:
I mean the code from right before this pull request.

ThomasZecha:
I got

ERROR: /home/tommy/repos/IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py:202: TypeError: 'module' object is not callable
  /home/tommy/repos/IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py:202
  /home/tommy/repos/IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py:229
  /home/tommy/repos/IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/tech/pymacros/autorun.lym:8 (class TypeError)
ERROR: ../../../../repos/klayout/src/db/db/gsiDeclDbCell.cc,157,cell != 0
ERROR: RuntimeError: Internal error: ../../../../repos/klayout/src/db/db/gsiDeclDbCell.cc:157 cell != 0 was not true in DCellInstArray.__init__
  pycell_test.py:52
  
  
[nmos.zip](https://github.com/user-attachments/files/17095726/nmos.zip)
with the attached PyCell

akrinke:
Thank you! I could fix the problem by changing line 29 in `sg13g2_pycell_lib/__init__.py` to:
```python
from pypreprocessor.pypreprocessor import preprocessor as preProcessor
```

In addition, I had to add `import imp` to `pypreprocessor/__init__.py`.

If you agree, I will create pull requests for it.

ThomasZecha:
Please create a PR, I will test it than.

akrinke:
The second PR is #210.

PR #215, opened 2024-09-30
iho-sg13g2: libs.ref: sg13g2_io: verilog: Fix specify syntax

### LINKED ISSUE #208: [bug] LRM specify block delay path restrictions
Hi, All!

I noticed that in [verilog cell file](https://github.com/IHP-GmbH/IHP-Open-PDK/blob/5a42d03194e8c98558f4e34538338a60550f89b9/ihp-sg13g2/libs.ref/sg13g2_io/verilog/sg13g2_io.v#L75) there are a lot of `specify` constraints which are buggy due to LRM:

> 30.4.1 Module path restrictions
> Module paths have the following restrictions:
> —The module path source shall be a net that is connected to a module input port or inout port.
> —The module path destination shall be a net or variable that is connected to a module output port or
> inout port.

Path can't start at output port due to LRM 2017 and it's earliest versions. So Does it correct to specify such constraints in that manner or it's bug in provided cells?

I use the [`slang`](https://github.com/MikePopoloski/slang/)  tool for elaboration of designs from https://github.com/IHP-GmbH/IHP-Open-PDK/tree/main . With its help I discovered this

dnltz:
The input and output definition in the Verilog specify block were positioned in the wrong order (output to input). Change and also implement the tri-state better.

Fixes #208

### DIFF
--- a/ihp-sg13g2/libs.ref/sg13g2_io/verilog/sg13g2_io.v
+++ b/ihp-sg13g2/libs.ref/sg13g2_io/verilog/sg13g2_io.v
@@ -72,7 +72,7 @@ module sg13g2_IOPadIn (pad, p2c);
 
 	// Timing
 	specify
-		(p2c => pad) = 0;
+		(pad => p2c) = 0;
 	endspecify
 endmodule
 `endcelldefine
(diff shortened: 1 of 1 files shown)


dnltz:
I also ran this through the online slang tool and it didn't report any errors.

PR #220, opened 2024-10-14
Fixes to the tech layer map used by klayout for DEF -> GDS

smunaut:


KrzysztofHerman:
Thank you for this PR. I have tested is elaborating `ricscv32i` design using actual and proposed version of the map file. After streaming GDS and making XOR (using klayout) I found no difference.  Could you please  provide/reference some testcase where we could actually see the effects ?

smunaut:
Note that to see the difference you need to change the "stipple" for the `drawing` layer because in the default klayout layer properties the difference is near invisible.

2024-10-17_1918x1178_scrot_002
2024-10-17_1918x1178_scrot

[pin_issue.zip](https://github.com/user-attachments/files/17411744/pin_issue.zip)

smunaut:
BTW I think the issue is not usually visible in ORFS because the pins created by default are the same width as the traces and might have their anchor point on the edge rather than in the center.

KrzysztofHerman:
The patch provides changes in  `sg13g2.map` file which makes that the pins area also have `drawing` datatype in the GDS and not just `pin`

PR #244, opened 2024-10-30
libs.tech/klayout: Fix the seal ring pcell EdgeSeal/boundary

smunaut:
The `w` and `l` are swapped so if you generate a non-square Seal Ring, then that layer is rotated 90 deg vs the rest of the ring.


smunaut:
See https://i.imgur.com/CWwRHjX.png for an illustration of the issue.

dnltz (approved):
True, as defined in `defineParamSepcs`
```
        specs('l', defL, 'Length(X-Axis)')
        specs('w', defW, 'Width(Y-Axis)')
```

PR #256, opened 2024-11-10
Add more Cnt rules

dnltz:
More Cnt rules and a test GDS file to trigger violations. The file only contains Cont violations but can be extended in the future!

layout is drawn after the layout rules with each rule violating the requirements.

image


dnltz:
@akrinke I figured out Cnt.g1 generates no violation. Can you take a look why it doesn't?

KrzysztofHerman:
@dnltz this `Cnt.g1` rule is still not implemented in the DRC deck. 
The full list of not implemented rules is here:
https://github.com/IHP-GmbH/IHP-Open-PDK/blob/main/ihp-sg13g2/libs.tech/klayout/tech/drc/MissingRules_maximal.md

akrinke:
We have implemented new DRC rules in PR #259, including `Cnt.c`, `Cnt.g1` and `Cnt.g2` that are part of this pull request. Maybe you could remove the rule implementations from this PR, but keep the test infrastructure?

dnltz:
@sergeiandreyev ah, perfect they already exists. Thanks for pointing them out.

@akrinke Yes I can do that :)

sergeiandreyev:
Hi @dnltz, so what do you think, can we close this PR? since the updates are provided by @akrinke

PR #255, opened 2024-11-08
Basic filler scripts for ihp-sg13g2 open PDK

KrzysztofHerman:
This PR contains basic filler scripts for the PDK.

The filling process can be run from the command line using a python script:
```
klayout -n sg13g2 -zz -r $PDK_ROOT/$PDK/libs.tech/klayout/tech/scripts/filler.py -rd output_file=./design_filled.gds design_nofill.gds   
```
It is mandatory to have a `sealring` in the design. 

In order to control metal filler densities user can modify the following scripts:

```
$PDK_ROOT/$PDK/libs.tech/klayout/tech/scripts/macros/sg13g2_filler_Metal.lym  
$PDK_ROOT/$PDK/libs.tech/klayout/tech/scripts/macros/sg13g2_filler_TopMetal.lym
``` 
where the `width`, `length` and also the `distance` between the filler cells can be altered.  Please respect the DRC rules related to the sizes and separation of the filler cells (sections `5.18` and `5.23` of the `SG13G2_Layout_Rules_os.pdf` document). 

While running `minimal DRC` in klayout one can expect ` AFil.g2`, `M1-5Fil.h`  errors related to the area outside the design, where the densities are calculated. The errors can be waived. 

Special thanks to `Daniel Schultz` for providing the scripts.

PR #260, opened 2024-11-12
Filler scripts hotfix to the main

KrzysztofHerman:
This PR integrates filler scripts used for Nov'204 tape-out. The PR #255 will be removed.


ihp-sg13g2/libs.tech/klayout/tech/macros/sg13g2_density_report.lym
sergeiandreyev, line 1: should be changed to `EdgeSeal.boundary`

PR #269, opened 2024-11-18
new library concept for corner/statistic/mismatch simulation

dwarning:
Additional to corner and statistically simulation allows this extension local mismatch simulation with ngspice monte-carlo approach. This is valid for hbt, lv-nmos and hv-nmos. Passive elements are unchanged and have to be revised.
The corner lib's have extended entries for typical, corner and mismatch and combination block's which the user have to specify. No more specifications of mm_ok and mc_ok on schematic/netlist level because of their meaning of number of std-deviation this was error-prone. They are setted internally.
Statistical parameter are untouched. Their meaning in compare to corner have to be investigated.
Mismatch parameter are derived from few entries in old lib and compared to empirical values. These values have to be verified. The mc varied parameter (and partly area dependent) go to instance line of the psp model.

Ngspice test cases will be provided.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
@@ -58,6 +58,7 @@ moduleNames = [
         'inductor3_code',
         'dantenna_code',
         'dpantenna_code',
+        'via_stack_code',
         'ptap1_code',
         'bondpad_code'
 ]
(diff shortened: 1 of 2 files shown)


ihp-sg13g2/libs.tech/ngspice/models/sg13g2_hbt_mod.lib
sergeiandreyev, line 1: please update to the latest version, it seems you're using not up-to-date models (the pin order was fixed Sep 30)
or did you make the change intentionally?

PR #331, opened 2025-01-14
provide a switch for vbic NQS effect

### LINKED ISSUE #309: Signal Amplitude Depends on Simulation Duration in Ngspice 43 Transient Simulation
## Environment
Ngspice 43
Qucs-s 24.3

I am encountering an unexpected behavior with Ngspice version 43 during the transient simulation. The issue is related to the amplitude of the signal, which seems to depend on the simulation duration. This problem does not occur in Ngspice version 42. Below is a detailed explanation:

Input signal: A squared current source:
Amplitude: 100 uA
Duration of High Pulse: 20 ps
Duration of Low Pulse: 20 ps
Rise Time: 20 ps
Fall Time: 20 ps

The screenshots attached with the description were the transient simulated output of an amplifier designed using an SG132 device.

## Expected Behavior
The expected behavior is created using **Ngpice 42**.

With 3 ns simulation duration, The signal amplitude:
image

With 2 ns simulation duration, The signal amplitude:
image

## Actual Behavior
The Actual behavior is created using **Ngspice 43.**

With a 2 ns simulation duration: The signal amplitude is very low:
Screenshot_7

Witha  3 ns simulation duration: The signal amplitude is very high:
Screenshot_1

## Steps to Reproduce the Problem
1. Create a simple circuit that generates a square wave signal with long rise and fall times. 
2. Run a transient simulation with different simulation durations (e.g., 2 ns and 3 ns).
3. Observe the output signal's amplitude.

dwarning:
Fixes #309

The switch sw_nqs is a parameter of the npn13G2 subckt. Default is 0 - setting it to 1 enables the NQS model in vbic.
In schematic an appropriate symbol is needed.

### DIFF
--- a/ihp-sg13g2/libs.tech/ngspice/models/sg13g2_hbt_mod.lib
+++ b/ihp-sg13g2/libs.tech/ngspice/models/sg13g2_hbt_mod.lib
@@ -43,6 +43,7 @@
 +Ny=1 le=0.96e-6 we=0.12e-6
 +El=le*1e6
 +selft=0
++sw_nqs=0
 
 Qnpn13G2 c b e s1 t npn13G2_NX_vbic dtemp=dtemp m=1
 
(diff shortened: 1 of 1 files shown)


dwarning:
Because of the discussed merge confusion this PR must deleted and applied new afterwards the mismatch PR is applied.

sergeiandreyev:
Hi @dwarning, as I noted during the meeting, we agreed to change the default setting of this parameter to `1`, am I right?
And to write a recommendation for the designers to switch it off if required.

dwarning:
@sergeiandreyev I would prefer to make it vice versus. Switch off as default in ngspice lib.   In case we find a solution for converging time domain simulations by improving the model we can switch NQS on.

sergeiandreyev (approved):

PR #335, opened 2025-01-17
Simple inductor PCell

adrienluitot:
This PR add a new kind of inductor PCell layout. It's not a differential one like the inductor2/3 currently proposed by the PDK. It allows a lower serial resistance and thus a higher Q (at least around few GHz), measured with OpenEMS. 

I haven't tried to LVS, I don't know how it will react.
The code is highly based on the inductor type 2 code.

I would like to add callbacks on the parameters, I could add them before the PR is merged. But I don't know how to implement them, I haven't found an example in current PCells. I could also add them in an future PR, when I know how it works.

The inductor looks like that in KLayout:

simple_inductor

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
@@ -56,6 +56,7 @@ moduleNames = [
         'npn13G2V_code',
         'inductor2_code',
         'inductor3_code',
+        'simple_inductor_code',
         'dantenna_code',
         'dpantenna_code',
         'via_stack_code',
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/ihp/simple_inductor_code.py
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/ihp/simple_inductor_code.py
@@ -1,0 +1,269 @@
+########################################################################
+#
+# Copyright 2024 IHP PDK Authors
+#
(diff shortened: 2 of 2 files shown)


VolkerMuehlhaus:
Removing the crossovers results in higher Q indeed, if you don't need symmetric layout. 

Another possible layout modification to increase Q, which also work for symmetric layouts: You can stack TopMetal1 and TopMetal2 for the coil turns whereever possible. Place a few vias every 90° to connect the metals.

adrienluitot:
Thank you for your feedback, it's indeed for non symmetrical layouts.

Stacking is a good idea too, why not exploring it some day.

sergeiandreyev:
Hi @atorkmabrains, @FaragElsayed2, could you please take a look if the LVS will work with this new inductor device or there will be some change needed to be compatible with OpenPDK?

FaragElsayed2:
Hi @sergeiandreyev 

No, LVS only supports inductor2* & inductor3* devices. The text serves as an identifier for recognizing inductor devices and their pins.

https://github.com/mabrains/IHP-Open-PDK/blob/4c6508d03a3078b21c737d04fae5dccec9aa590f/ihp-sg13g2/libs.tech/klayout/tech/lvs/rule_decks/ind_derivations.lvs#L36

adrienluitot:
Okay, thank you for your feedbacks.
I don’t have a lot of time for the moment for it, but as soon as I have, I will give a look at the LVS. And add some features (maybe stacked metal, pins I opposite directions, vias count… )

I would like to add callbacks too, if anyone knows how to do some.

I also need to fix the DCO.

PR #338, opened 2025-01-21
Res r3 model issue #318

### LINKED ISSUE #318: Noise Analysis Issue with PDK Resistor in Transimpedance Amplifier Design
I am designing a Transimpedance Amplifier (TIA) using the PDK and have encountered a significant issue during noise analysis. Below are the details:

Problem Statement:
Noise performance is critical for my TIA design. I initially tested the circuit using ideal resistors and inductors, and the noise levels were in the expected range of **pA/√Hz**. However, upon replacing the ideal resistor with the PDK resistor, the noise levels changed drastically, increasing from **pA/√Hz** to **nA/√Hz**. This discrepancy raises concerns about the accuracy of noise analysis with the PDK resistor.

Simulation Setup:

Simulator: **Ngspice 44**
Schematic Editor: **Qucs-s 24.4.1**

Circuit Details: The design uses a combination of ideal and PDK components. The issue arises specifically when substituting ideal resistors with PDK resistors.

Steps to Reproduce:
Design a TIA circuit in Qucs-s using Ngspice as the simulator.
Perform noise analysis using ideal resistors and note the results (pA/√Hz range).
Replace the ideal resistor with a PDK resistor.
Rerun noise analysis and observe the noise level (nA/√Hz range).

Observed Behavior:
The noise levels jump significantly from the **practical range of pA/√Hz** to an **unrealistic nA/√Hz** after replacing the ideal resistor with the PDK resistor.
There is uncertainty about whether the noise analysis is functioning correctly with the PDK resistor.

Expected Behavior:
The noise levels should remain consistent with practical expectations (in the pA/√Hz range) when using the PDK resistor, assuming it is modeled correctly.
Noise analysis should work properly with PDK components, providing reliable results.

Additional Context:
This issue might indicate an error in the noise model of the PDK resistor or a compatibility issue with Ngspice.
If the noise analysis for the PDK resistor is not supported or requires specific configuration, additional clarification in the documentation would be helpful.

Request for Support:
Could you please investigate if this issue is related to the PDK resistor model or its compatibility with noise analysis in Ngspice? Any guidance on resolving this problem or verifying the noise analysis would be greatly appreciated.

Attachments:
[CE_Buffer_3_noise_analysis.zip](https://github.com/user-attachments/files/18328354/CE_Buffer_3_noise_analysis.zip)
image
image

dwarning:
This PR fixes #318 and by using the r3_cmc Verilog-A model is now compatible with the model used in the commercial pdk.
Attached document show very few simulation examples also for noise simulation amongst others:

[sg13_res_r3_model.pdf](https://github.com/user-attachments/files/18488277/sg13_res_r3_model.pdf)

Because of using a Verilog-A model reorganization of directory structure was needed and adms or openvaf is needed respectively used simulator.


melikyazici:
The *rshspec* parameter for *rppd* is not explicitly defined in the *resistors_mod.lib* and *resistors_mod_mismatch.lib* files under the *rppd* subcircuit model, similar to *rsil* and *rhigh*.

dwarning:
It is not wrong because the rsh_rppd (and the others) are use and defined in cornerRES.lib.
But you are right - the rshspec parameter is not used anymore. Both line should be removed (in rsil and rhigh subct).
Would be cleaner.
Or is this subckt parameter user from symbol side under gui?
Thank you.

hpretl:
Looking at the sheer number of changed files (incl. a change of the folder structure) this PR is hefty, but switching to the R3-CMC model is excellent to have. I think we should also merge soon, before this triggers even more complications later.

@sergeiandreyev @KrzysztofHerman What do you think? @dwarning Do you think we are ready to merge?

sergeiandreyev:
Hi @hpretl, this PR went through internal IHP review and as I understand it is ready for merge, I'll double check tomorrow

sergeiandreyev (approved):

PR #354, opened 2025-02-11
Xschem npn13G2v symbols updated 

KrzysztofHerman:
npn13G2v symbols updated in order to support `El` parameter. Also DRC rules updates for xschem


sergeiandreyev:
thank you! could you please update the following:
- default `El` to `1`
btw, we have a disconnect with the Nx limits..

sergeiandreyev (approved):

PR #377, opened 2025-03-05
Make Latch-up rules optional

dnltz:
Allow to disable the Latch-up rules via a KLayout parameter to not check against them.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/drc/sg13g2_maximal.lydrc
+++ b/ihp-sg13g2/libs.tech/klayout/tech/drc/sg13g2_maximal.lydrc
@@ -131,6 +131,11 @@ if defined? $sanityRules
 else
     $sanityRules = true
 end
+if defined? $latchUpRules
+    $latchUpRules = $latchUpRules.to_s.downcase == "true"
+else
+    $latchUpRules = false
+end
 
 class AbuttingEdges &lt; RBA::EdgePairToEdgeOperator
     def initialize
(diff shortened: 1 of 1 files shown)


dnltz:
@akrinke FYI

dnltz:
@KrzysztofHerman I think we can close that with #379

KrzysztofHerman:
@dnltz thank you for the contribution. I have checked this option and IMHO the logic should be inverted, right now if you put:
```
klayout-s -zz -s latchup_violation.gds -rd noLatchUpRules=false -r $PDK_ROOT/ihp-sg13g2/libs.tech/klayout/tech/drc/sg13g2_maximal.lydrc 
``` 
the rules LatchUp rules are checked. If you set it to `true` you disable the check. It is inconsistent with other checks. 
Could you please fix this. I guess it is in the line 3049

akrinke:
@dnltz Thank you, I think having the option to disable latchup checks is a good idea!

@KrzysztofHerman Currently, we have two types of options: `offGrid`, `filler`, `density`, and `sanityRules` are enabled (`true`) by default, while `noRecommendedRules` is `false` by default. But I agree, `latchup` with a default value of `true` makes more sense.

KrzysztofHerman:
@dnltz and @akrinke finally the goal is to have these options in the SG13G2 menu item where the checks extends minimal set of rules by the one enclosed by `if` statement,

akrinke:
@KrzysztofHerman Yes, I agree. I will implement it.

dnltz:
> @dnltz thank you for the contribution. I have checked this option and IMHO the logic should be inverted, right now if you put:
> 
> ```
> klayout-s -zz -s latchup_violation.gds -rd noLatchUpRules=false -r $PDK_ROOT/ihp-sg13g2/libs.tech/klayout/tech/drc/sg13g2_maximal.lydrc 
> ```
> 
> the rules LatchUp rules are checked. If you set it to `true` you disable the check. It is inconsistent with other checks. Could you please fix this. I guess it is in the line 3049

@KrzysztofHerman  - As @akrinke already said, it's a copy of `noRecommendedRules` and you need to actively disable the latch-up rules. I think latch-up rules should be enabled by default. 

However, I can invert the flag if you want to.

akrinke:
@dnltz I agree that latchup rules should be enabled by default – this should not change. So I recomment to name the variable `latchup` with a default value of `true`.

sergeiandreyev:
> > @dnltz thank you for the contribution. I have checked this option and IMHO the logic should be inverted, right now if you put:
> > ```
> > klayout-s -zz -s latchup_violation.gds -rd noLatchUpRules=false -r $PDK_ROOT/ihp-sg13g2/libs.tech/klayout/tech/drc/sg13g2_maximal.lydrc 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the rules LatchUp rules are checked. If you set it to `true` you disable the check. It is inconsistent with other checks. Could you please fix this. I guess it is in the line 3049
> 
> @KrzysztofHerman - As @akrinke already said, it's a copy of `noRecommendedRules` and you need to actively disable the latch-up rules. I think latch-up rules should be enabled by default.
> 
> However, I can invert the flag if you want to.

`noRecommendedRules` is a very weird way of implementing a boolean switch, that's historically coming from the internal PDK rules implementation
I'll strongly suggest updating/aligning it to other switches

sergeiandreyev:
> @dnltz Thank you, I think having the option to disable latchup checks is a good idea!
> 
> @KrzysztofHerman Currently, we have two types of options: `offGrid`, `filler`, `density`, and `sanityRules` are enabled (`true`) by default, while `noRecommendedRules` is `false` by default. But I agree, `latchup` with a default value of `true` makes more sense.

for all these different categories, we should have a clear documentation (explanation, usage guide) on when what to set, why the defaults are like this, etc.
something is already written in DRM, but certainly is not enough, f.e. there is no any mentioning of what are these `sanity` rules
@akrinke, do you know some details on the `sanity` checks apart from the rules list that are included in this category?

akrinke:
@sergeiandreyev Sanity checks include checks for forbidden layers (see Sec. 3.2 in the layout rules PDF) and checks if all pin areas are covered by drawing (see Sec. 7.4). Should I remove this flag and thus prevent their deactivation?

sergeiandreyev:
let's wait - I'm currently discussing this internally

KrzysztofHerman:
@dnltz could you please update the PR by

1. changing the name from `noLatchUpRules` to `LatchUpRules`
2. inverting the behavior of the true/false switches in the command line (the rules should be evaluated when `LatchUpRules=true`)

The rules evaluation should be enabled by default

dnltz:
Changes `noLatchUpRules` to `latchUpRules` and inverted the logic. `latchUpRules=true` is now required.

akrinke:
@dnltz I think the rules should be enabled by default. The inversion referred to the change from `noLatchUpRules` to `latchUpRules`.

akrinke:
Sorry for the confusion, I meant the PR should be changed to enable the rules by default. The latest version of this PR (the one that has been merged an hour ago) **disables** latch-up rules by default.

sergeiandreyev:
Hi @akrinke, yes, I fixed this in my post-PR commit: https://github.com/IHP-GmbH/IHP-Open-PDK/commit/3b393f31a3d6f66b995100595019e6e096ccff05

akrinke:
@sergeiandreyev Perfect – I had missed this commit.

PR #382, opened 2025-03-06
IO cell GDS updated with latch up errors fix

b10346:
IO cell GDS updated with latch up errors fix


dnltz:
image

Corner cells are now rotated. Is that intended? Probably easy fix in the pad configuration but just wanna make sure it's for purpose?

b10346:
Orientation of Corner cell is aligned with another layout database of the library. Is it a big effort to make changes in a design?

dnltz:
I think there is a mismatch between files now. This is how it looks like in OpenROAD

image

and that's klayout (with only some layer to actually the corner cells)

image

dnltz:
you can see they are mirrored in the GDS file now.

sergeiandreyev:
yep, I confirm the rotation..
old version:
image
new version:
image

b10346:
Fix with Corner cell rotation will be soon.

b10346:
cancelled due to corner rotation issue

PR #409, opened 2025-03-25
Corrections to fill pattern generation and density checks in magic.

### LINKED ISSUE #408: Fill pattern generation in Magic has multiple issues.
I am entering this issue so that it is recorded in the issue tracker and not just a pull request.

The pattern generation parameters in the magic tech file has multiple problems:

1. The density for all metal layers and diffusion is too high, approximately 70% for most layers, while a more reasonable target is 50% or slightly below.
2. The resizing used to eliminate fill shapes that are clipped by keep-out areas is incorrect in a few cases, resulting in some fill shapes either non-rectangular or smaller than allowed.
3. The poly fill is not aligned with the diffusion fill, and in fact is being suppressed by the diffusion fill, resulting in too-low poly density.
4. The density fill script needs rewriting, to handle GDS files as well as .mag files, to automatically find the .magicrc startup script, and to make use of compressed GDS for the pattern fill files, among other things.
5. The density check script has incorrect calculations for the area as counted by whole tiles, having the original 100 count for a 10 x 10 stepped area, which was carried over from the original script used as the source for this one, whereas the steps were changed in this file to step by half the width of a tile, so the total tile area is 4 for a 2 x 2 stepped area.
6. The density check script has incorrect calculations for pro-rating the density of partial tiles on the right and top sides of a layout.

These items are being addressed and will result in a pull request.

RTimothyEdwards:
Made a number of changes in support of pattern fill generation and density checks.  Modified the tech file parameters for the generated fill to center the fill-related density on 50% fill. Corrected an issue with failure to generate poly fill on top of diffusion fill, and then to ensure that the poly fill shape is centered on the diffusion fill shape.  Added different position offsets to each of the metal fill layers so they are not stacked directly on top of one another.  Corrected the expressions for calculating the pro-rated densities on the right and top edge of a layout.  Added some documentation of pattern fill and density checks to the README file, and updated the RELEASE_NOTES file.

Fixes #408 

- [X] Tests pass
- [X] Appropriate changes to README are included in PR

Signed-off-by: R. Timothy Edwards <tim@opencircuitdesign.com>

### DIFF
--- a/ihp-sg13g2/libs.tech/magic/README.md
+++ b/ihp-sg13g2/libs.tech/magic/README.md
@@ -251,3 +251,39 @@ known sheet resistance values.
 |-------------|-----------------------------------------------------|
 | hvpvaractor | parasitic device formed at the ends of SVaricap     |
 | fillfet     | device formed at the junction of fill diff and poly |
+
+## Fill pattern generation
+
+Fill pattern generation is implemented as an alternative "cif ostyle"
+in the magic tech file.  It is only marginally useful in an
+interactive layout session;  generally, it is meant to be used with
+the standalone python script "generate_fill.py".  Running the python
+script (run without arguments to get a usage statement) on a magic or
+GDS database of a chip top level (including seal ring) will produce
+a compressed GDS file containing a overlay cell with the fill
+patterns.  The fill pattern GDS must be integrated with the chip top
+level (the script does not do this) by including the fill patterns
+into the top level layout, with the origins aligned.
+
+If the intention is to pre-fill an area of layout, magic can do this
+with the command "cif ostyle patternfill()", placing the cursor box
+around the area to be filled, and typing, e.g., "cif paint MET1FILL
+met1fill".  Each fill layer must be generated individually.
+
+## Auxiliary DRC checks
+
+Not all DRC rules are implemented directly by magic's DRC checker.
+The two main exceptions are density and antenna rules.  Density rules
+are only meaningful if done on an entire top-level layout, after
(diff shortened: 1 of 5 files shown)


RTimothyEdwards:
Please hold off on merging this just yet, since after clarification from Krzysztof about the definition of the outer boundary where fill shapes should be generated, I will need to make a correction to the fill generation script.

RTimothyEdwards:
@KrzysztofHerman , @sergeiandreyev :  I have made the additional changes and this pull request is now ready for review.

RTimothyEdwards:
I have put this pull request in draft mode;  I added one more change because Sylvain found a recurring DRC error that was not present in the full-chip example that I ran to validate the pattern fill (this error has been fixed by the 3rd commit).  I am still waiting on a report on additional errors that seem like false positives, but I want to be certain before I take this pull request out of draft mode again.

smunaut:
The `GFil.d.pSD` errors are real.
All the other `MnFil.a1` violations are bugs in the DRC AFAICT.

RTimothyEdwards:
@smunaut :  Okay, thanks.  The last commit should fix the `GFil.d.pSD` errors.  Since the rest of them appear to be false positives, I will take this pull request out of draft mode.

smunaut:
Ping on this ? We used it for TT tape out successfully ...and it's a definite improvement vs what's currently in the repo.

KrzysztofHerman:
@RTimothyEdwards @smunaut relying on the opinion of Sylvain.

PR #426, opened 2025-04-03
libs.tech/sg13g2_io: Add nofill zones over IO cells

### LINKED ISSUE #423: Enhancement Req: Add `nofill` zones in IO cells
The IO cells are very dense in `Metal` and `Activ`, so for most layers it doesn't make sense to have fill in them. And even if the fill algo can find some small nook to stuff a tiny fill shape in between two giant power rails, I don't think it's really desirable.

What I propose is to add `nofill` overlays on the IO cells for all fill layers except for:
* `Metal2` : There is no big rail on this layer in the cells, so you actually want `Metal2` fill
* `GatPoly` : If you have a `Activ:nofill`, there won't be any `GatPoly` fill anyway and there is a `GFil.i` DRC rule that'd be violated.

For the zone to be covered, I used the same area as `prBoundary` except for the `Corner` cell where I cut one of the corner.

Attached are : 
 * A `gdstk` script that does the modification suggested above.
 * A GDS modified with the above script

[ihp_io_nofill.py.gz](https://github.com/user-attachments/files/19551616/ihp_io_nofill.py.gz)
[sg13g2_io_nofill.gds.gz](https://github.com/user-attachments/files/19551615/sg13g2_io_nofill.gds.gz)

Also an example picture with `Metal5` and the `nofill` area.

Image

Of course the script and GDS are only for illustration, if the cells have some other source format / script to generate them, that's where the modification should be made, that's why I didn't open a PR but raised an issue.

smunaut:
Add nofill overlays on the IO cells for all fill layers except for:

 * Metal2 : There are no big rails on this layer in the cells, so you actually want Metal2 fill

 * GatPoly : If you have a Activ:nofill, there won't be any GatPoly fill anyway and there is a GFil.i DRC rule that'd be violated.

For the zone to be covered, the same area as prBoundary is used except for the Corner cell where one corner is cut.

Fixes #423


sergeiandreyev:
@smunaut, thank you for the contribution! but I would like not to mix PDK library deliverables (IO cells) with the post-P&R pre-submission stage (fillers/slits generation to achieve desired density), the other thing the layouts are (should be) automatically generated by PDKMaster, which means every customization has to be backported to PDKMaster code for consistency
so, we should have an internal discussion on this, hope this is fine for you..
btw, we added `NoFillerStack` Pycell a couple of months ago to the `dev` branch

PR #427, opened 2025-04-03
Updating mim derivations for SG13G2-LVS

### LINKED ISSUE #347: LVS behaviour with parallel MIM Capacitors
When I stick 2 MIM caps with through Metal 5 and TopMetal 1 to use them in parallel, the LVS won't detect the capacitors at all. (Case 1)
If I space them, the LVS will detect 2 independent MIM caps. (Case 2)
And if keep the space, but fill it with Metal 5 and TopMetal 1 to connect them as if they were stuck, the LVS detect 1 MIM cap with a multiplier=2. (Case 3)

For the case 2 and 3, the space can be as small as 0.005um.

In the end, what I would expect, is a Layout like in case 1, but with LVS results like in case 3.
I don't really understand why the LVS results between case 1 and 3 are so different.

Also, maybe I'm not doing the parallelisation properly?

---
### Case 1
Layout:

LVS: <empty>

---
### Case 2
Layout:

LVS:
Image

---
### Case 3
Layout:

LVS:

Image

FaragElsayed2:
### Updating mim derivations for SG13G2-LVS

Fixes #347


FaragElsayed2:
@sergeiandreyev Could you please re-run the git actions for the failed jobs? I believe they failed due to a connection issue.

sergeiandreyev:
> @sergeiandreyev Could you please re-run the git actions for the failed jobs? I believe they failed due to a connection issue.

Hi @FaragElsayed2, yes, pushed the button
@KrzysztofHerman, haven't we fixed this issue with KLayout download timeout before? if yes, then the solution is not working as expected; if not, then let's discuss how to fix it

PR #441, opened 2025-04-09
#435 workaround for statistical simulations - provided by user smunaut

### LINKED ISSUE #435: `tt_stat` models are picking new random parameters per device instance
So what is expected when running a `tt_stat` sims, is that random device parameters are picked once (per device type) and then applied to all instances of that devices in the design.

This is not what happening. Instead each instance gets different random parameters which as you can imagine is a bit unrealistic because the spread from `tt_stat` is quite wide and using random parameters in current mirrors or differential pairs leads to "interesting" results.

The way the models pick the parameters is using `.param xxxx = gauss(....)` and if you believe the `ngspice` manual, theses should be evaluated once at startup and the same value used everywhere, which is exactly what you'd want here.

Unfortunately, the manual is wrong ... I've opened a thread on the `ngspice` mailing list ( https://sourceforge.net/p/ngspice/discussion/133842/thread/c13bbd10c6/ ) but the gist of it is that there seem to be a deliberate, undocumented behavior that any `param` that has a statistical distribution function in it is re-evaluated at each use.

That obviously wreaks havoc here the way the models are written ... it basically creates a mismatch simulation but with a much greater spread of parameters than you'd want for a mismatch sim :/ 

One work around I've tried is instead of : 
```
.param xxx = gauss(...)
```

is to use :

```
.param rnd_xxx = gauss(...)
.param xxx = rnd_xxx
```

Then `xxx` will actually have a stable value.

This is an example patch that applies that work around to the models here.
[ihp_stat_workaround.diff.gz](https://github.com/user-attachments/files/19628464/ihp_stat_workaround.diff.gz)

Note that this is for testing/evaluation only. I only checked the `lv nmos` with it applied and I'd also like to see how the `ngspice` thread before applying anything, but I'm opening the issue here to keep track / not forget about it.

dwarning:
Fixes #435

This patch fixes the weakness of ngspice to handle statistical variations in model parameters which are include inside subckts.
Better solution would be to prevent usage of instance parameters in model parameter calculation and place the model definition outside the subckt.


KrzysztofHerman (approved):

KrzysztofHerman:
Here I add a MOSlv testcase:

```
**.subckt mc_mode_examle
I0 net1 Vgs1 10u
XM1 Vgs1 Vgs1 GND GND sg13_lv_nmos w=1.0u l=0.45u ng=1 m=1
I1 net1 Vgs2 10u
XM2 Vgs2 Vgs2 GND GND sg13_lv_nmos w=1.0u l=0.45u ng=1 m=1
V1 net1 GND 1.2
**** begin user architecture code

.lib cornerMOSlv.lib mos_tt_stat


.param temp=27

.control

let mc_runs = 10
let run = 0
set curplot=new
set scratch=$curplot
setplot $scratch
let vg1=unitvec(mc_runs)
let vg2=unitvec(mc_runs)

***************** LOOP *********************
dowhile run < mc_runs

*dc Vds 0 3 0.01
op
set run=$&run
set dt=$curplot
setplot $scratch
let out1{$run}={$dt}.Vgs1
let Vg1[run]={$dt}.Vgs1
let out21{$run}={$dt}.Vgs2
let Vg2[run]={$dt}.Vgs2
setplot $dt
reset
let run=run+1
end
***************** LOOP *********************


print {$scratch}.vg1
print {$scratch}.vg2

.endc

**** end user architecture code
**.ends
.GLOBAL GND
.end

```

after executing one should get the same values of the `Vgs` voltage for each run

```
                                   Anonymous
                                   unknown  Tue May  6 13:21:14  2025
--------------------------------------------------------------------------------
Index   unknown1.vg1    
--------------------------------------------------------------------------------
0       4.555998e-01
1       4.042812e-01
2       4.878495e-01
3       4.939795e-01
4       4.278322e-01
5       4.575620e-01
6       5.163563e-01
7       4.417215e-01
8       5.180368e-01
9       5.151536e-01
                                   Anonymous
                                   unknown  Tue May  6 13:21:14  2025
--------------------------------------------------------------------------------
Index   unknown1.vg2    
--------------------------------------------------------------------------------
0       4.555998e-01
1       4.042812e-01
2       4.878495e-01
3       4.939795e-01
4       4.278322e-01
5       4.575620e-01
6       5.163563e-01
7       4.417215e-01
8       5.180368e-01
9       5.151536e-01
Note: Simulation executed from .control section 

```

ihp-sg13g2/libs.tech/ngspice/models/resistors_stat.lib
hpretl, line 1: @KrzysztofHerman Why is the parameter `mc_ok` not used for the resistors? It looks like it is used in all other components.

KrzysztofHerman:
@hpretl @dwarning 

I took a look at the NGSPICE and HSPICE manuals and found, together with ChatGPT, the following:

`NGSPICE` summary:
| Function Call              | Description                                                                                                                               |
| -------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
| `agauss(nom, avar, sigma)` | Nominal value plus variation drawn from a Gaussian distribution with mean 0 and standard deviation `avar` (absolute), divided by `sigma`. |

`HSPICE` summary:

---

image

## 📘 Using `sigma` in `AGAUSS(nominal_val, abs_variation, sigma)`

The `AGAUSS` function introduces Gaussian (normal) variation around a nominal value. The `sigma` parameter controls how the **absolute variation** translates into the **standard deviation**.

---

### 🔧 Syntax

```spice
AGAUSS(nominal_val, abs_variation, sigma)
```

* **`nominal_val`** – The nominal (mean) value for simulation.
* **`abs_variation`** – Total variation range: `±abs_variation`.
* **`sigma`** – Defines the number of standard deviations (`σ`) that span the `abs_variation`.

---

### 📊 How `sigma` Works

The standard deviation used in the distribution is:

```
σ = abs_variation / sigma
```

This means:

* About **68.2%** of samples lie within ±1σ.
* About **95.4%** of samples lie within ±2σ.
* About **99.7%** of samples lie within ±3σ.

---

### ✅ Example

```spice
AGAUSS(1.0, 0.09, 3)
```

* `nominal_val` = 1.0
* `abs_variation` = ±0.09
* `sigma` = 3 → standard deviation = 0.09 / 3 = 0.03

The resulting distribution:

* Mean = 1.0
* Std. dev. = 0.03
* Most values (99.7%) will fall in the range **0.91 to 1.09**

---

### 📐 Summary Table

| Desired Confidence Interval | Typical `sigma` | Coverage Range          |   |
| --------------------------- | --------------- | ----------------------- | - |
| \~68%                       | 1               | ±1σ = ±abs\_variation/1 |   |
| \~95%                       | 2               | ±2σ = ±abs\_variation/2 |   |
| \~99.7%                     | 3               | ±3σ = ±abs\_variation/3 |   |

---

### Summary 
I believe that the switch `mc_ok` comes from the `SKY130` PDK, where it was used to enable/disable the MC simulation. 
In fact this parameter gives the designer pretty comprehensive handle to modify the confidence interval. 

@dwarning please take a look at the pp. 880 of this [manual](https://cseweb.ucsd.edu/classes/wi10/cse241a/assign/hspice_sa.pdf). Seems that `hspice` has the same behavior as `ngspice`.

dwarning:
To have same behavior as hspice was the initial intention of this implementation.

But again, to give sigma value free to the designer is an bad idea. This sigma/variation couple should come from process description and fixed for MC simulation. Same for mismatch simulation.

hpretl:
@dwarning Thanks for merging! But could you please look into this, I have a feeling this should be changed: https://github.com/IHP-GmbH/IHP-Open-PDK/pull/441#discussion_r2075486230

dwarning:
I agree, we should unify the statistical library - actual
resistor_mod_stat.lib differs.

But we should discuss with @Herman, Krzysztof
***@***.***> how we prevent that user can change the
sigma value erroneous.
My idea is to rename the parameter mc_ok to nsigma (number of sigmas) and
not to publish to the user.
He has to include the correct library entry on top level circuit if
statistical simulation is requested.

Am Mi., 7. Mai 2025 um 09:04 Uhr schrieb Harald Pretl <
***@***.***>:

KrzysztofHerman:
@hpretl 
We would like to have you feedback on this `mc_ok` switch and our way to change it. 
So we would like to get rid of this parameter at the top level. If you would like to enable/disable statistical simulation you just switch between the sections in  the `.lib` files ie. `mos_tt` vs. `mos_tt_stat`. 
Additionally in the  `*_stat.lib` files we would like to substitute `mc_ok` with a verbose and comprehensive name `num_sigmas`, which represents number of sigmas, by default set to 1. 

```
.param num_sigmas=1
* ngspice statistical parameters
.param mc_sg13g2_lv_nmos_vfbo     ='gauss(sg13g2_lv_nmos_vfbo_norm ,    0.0050, num_sigmas)'
```
Thus we will have a simple mechanism to enable/diable  MC analysis and additionally an advanced designer will have a handle to control the confidence interval.

hpretl:
@KrzysztofHerman 

I don't see why the designer should be able to change the sigma of the components' spreads. The variance of the devices should be as the process is, so I see exactly in simulation what the circuit would do in production. Then, when looking at simulated **circuit parameters**, I decide about which interval I accept for a parameter as spec, like +/- 3 sigma, or +/- 6 sigma, or whatever.

But this variation should only be discussed on circuit performance level. The individual components should always vary as they vary as in production.

hpretl:
@KrzysztofHerman @dwarning 

> My idea is to rename the parameter mc_ok to nsigma (number of sigmas) and
not to publish to the user.

Exactly, this is how it should be. Users should not be allowed to screw around with the component spreads.

dwarning:
OK, so it will be happen.
Thanks.

Issue #221, opened 2024-10-15
Separate KLayout DRC script and XML wrapper


mole99:
Currently, the DRC scripts are implemented using a `.lydrc` file, which is an XML wrapper around the DRC script: https://github.com/IHP-GmbH/IHP-Open-PDK/tree/dev/ihp-sg13g2/libs.tech/klayout/tech/drc

Instead, it would be great if the DRC script could be separated from the wrapper and simply included with `# %include`.
~~This would make it possible to run the DRC scripts from within a Ruby or Python script.~~ It is actually possible to run the .lydrc macro directly.

An example of how this works can be found in sky130. 
There is one [sky130A_mr.drc](https://github.com/efabless/sky130_klayout_pdk/blob/main/tech/sky130/drc/sky130A_mr.drc) DRC script that is included in several `.lydrc` files:

- [run_drc_beol.lydrc](https://github.com/efabless/sky130_klayout_pdk/blob/main/tech/sky130/macros/run_drc_beol.lydrc)
- [run_drc_feol.lydrc](https://github.com/efabless/sky130_klayout_pdk/blob/main/tech/sky130/macros/run_drc_feol.lydrc)
- [run_drc_full.lydrc](https://github.com/efabless/sky130_klayout_pdk/blob/main/tech/sky130/macros/run_drc_full.lydrc)

By setting the values of global variables, different DRC runsets can be enabled/disabled.

The same approach could be used to combine the `sg13g2_maximal` and `sg13g2_minimal` DRC scripts into a single file.

akrinke:
I agree that a single DRC script would be better because it avoids code duplication. However, due to the way we generate the scripts, some major changes are required. Therefore, it is not a priority for us at the moment.

In the end, it is not so different than for sky130: We have two `.lydrc` files, they have three. Besides, you can already run the `.lydrc` scripts from Ruby and Python:
```ruby
# ruby
RBA::Macro::new(path_to_lydrc).run()
```
Here is a full example:
```ruby
# DRC script configuration

unless $drc_script
  $drc_script = __dir__ + "/sg13g2_minimal.lydrc"
end

# END OF CONFIGURATION
######################

expanded_path = File.expand_path($drc_script)

start = Time.now
puts "Start: " + start.to_s

RBA::Macro::new(expanded_path).run()

finish = Time.now
puts "End:   " + finish.to_s

duration = finish - start
minutes = duration.to_i / 60
seconds = duration % 60

[...]

PR #444, opened 2025-04-10
ihp-sg13g2: libs.tech: klayout: Update Metal Filler Script

dnltz:
Make the metal filler script more agressive by adding a medium and small fill cell. Moreover, allow to define the distance between filler cells for each metal layer.

Default values are more conservative right now to not over-fill small designs. I see two possible improvements for later

1. Allow to load a YAML file with Mx distance values. That would allow to have different values also in batch mode
2. Add a large filler cell in case we change min or max width of metal filler cells. A four times bigger cell then `1.0x1.0` would be `5.26x5.26`

- [ ] Tests pass

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/macros/sg13g2_filler_Metal.lym
+++ b/ihp-sg13g2/libs.tech/klayout/tech/macros/sg13g2_filler_Metal.lym
@@ -111,70 +111,207 @@
 	LBE = source.input("157/0")
 	NoMetFiller = source.input("160/0")
 
+
+	# Create a line edit widget
+	#
+	# text - The default text for the line edit
+	#
+	# Returns a QCheckBox instance
+	def self.create_input_field(text)
+		line_edit = RBA::QLineEdit.new()
+		line_edit.setText(text)
+		line_edit
+	end
+
+	# Create buttons for the distance dialog
+	#
+	# dialog - The distances dialog
+	#
+	# Returns a QWidget instance containing the buttons
+	def self.create_buttons(dialog)
+		container = RBA::QWidget.new
+		layout = RBA::QHBoxLayout.new(container)
+
+		ok_button = RBA::QPushButton.new('OK')
+		ok_button.clicked.connect(Proc.new { dialog.accept })
+		cancel_button = RBA::QPushButton.new('Cancel')
+		cancel_button.clicked.connect(Proc.new { dialog.reject })
+
+		layout.addWidget(ok_button)
+		layout.addWidget(cancel_button)
+
+		container
+	end
+
+	# Create dialog to select custom distances between Mx filler
+	#
+	# defaults - Default distances to add to the input fields
+	def self.create_options_dialog(defaults)
+		dialog = RBA::QDialog.new
+		dialog.windowTitle = "Metal Fill Distances"
+		layout = RBA::QFormLayout.new(dialog)
+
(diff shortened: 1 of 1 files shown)


dnltz:
That's how the dialog looks like

image

dnltz:
@smunaut @mole99 FYI

mole99:
It's a good idea to make the fill script more customizable, especially since this is still necessary to fill certain layouts properly. But I'm not sure if all the GUI functionality should be packed into the `.lym` together with the DRC operations?

I would prefer to separate this: the DRC/filler script that does the filling and the KLayout macro that sets the inputs.

See for example in sky130, there we have three different macros to call the DRC script with different parameters: https://github.com/efabless/sky130_klayout_pdk/tree/main/tech/sky130/macros

Basically, just set the global parameters and then include the actual script using `# %include path/to/fill.drc`.

Yes I know that you check for `batch_mode`, but then this wouldn't even be necessary. You can just call the DRC/filler script directly in batch mode and set up the global parameters (or use sane defaults from the script).

See here for a similar request for the LVS script: https://github.com/IHP-GmbH/IHP-Open-PDK/issues/221

I hope this isn't too confusing 😄

dnltz:
> It's a good idea to make the fill script more customizable, especially since this is still necessary to fill certain layouts properly. But I'm not sure if all the GUI functionality should be packed into the `.lym` together with the DRC operations?
> 
> I would prefer to separate this: the DRC/filler script that does the filling and the KLayout macro that sets the inputs.
> 
> See for example in sky130, there we have three different macros to call the DRC script with different parameters: https://github.com/efabless/sky130_klayout_pdk/tree/main/tech/sky130/macros
> 
> Basically, just set the global parameters and then include the actual script using `# %include path/to/fill.drc`.
> 
> Yes I know that you check for `batch_mode`, but then this wouldn't even be necessary. You can just call the DRC/filler script directly in batch mode and set up the global parameters (or use sane defaults from the script).
> 
> See here for a similar request for the LVS script: #221
> 
> I hope this isn't too confusing 😄

I'm not quite sure what you mean with DRC script since this filler script doesn't run any DRCs (might be nice to directly check density violations afterwards).

My plan is to move the GUI functions into an include file and add YML support in the next iteration. I would also extent this feature to Activ/GatPoly and TM since someone might wanna adjust fill there too.

mole99:
The filler script is actually just a DRC script in KLayout, i.e. it runs on the same interpeter (`drc-dsl-xml`).

No, that's not actually what I meant. I meant that you should keep the GUI in the macro (`lym`) and move the actual filler script to a separate file, e.g. `fill.drc`. The `lym` then simply includes the filler script using `# %include path/to/fill.drc`. This makes it possible to run the filler script without the GUI and without any checking of `batch_mode`.

I think this is the proper way to do this.

sergeiandreyev:
but maybe not to slow down development, we can go step by step

mole99:
This is not much work at all:

1. Create a new file called `fill.drc` under e.g. `klayout/tech/drc/`
2. Move the fill specific code from `sg13g2_filler_Metal.lym` to `fill.drc`
3. Include `fill.drc` in `sg13g2_filler_Metal.lym` using `# %include ../drc/fill.drc`

The `sg13g2_filler_Metal.lym` macro then just holds the GUI code and the include, that's it.
By the way, this is just for code organisation. The `drc-dsl-xml` interpreter will still be used when called from the macro.

PR #453, opened 2025-04-15
mismatch selectable on instance level

dwarning:
This PR makes mismatch simulations more practical. The user has to select the devices which he want consider in mismatch simulation by setting mm_ok to 1 in the instance property box. This parameter will come into the subckt and will activate the gaussian function calculation in Monte Carlo runs.
The default for mm_ok is 0 - gaussian function is deactivated and delivers nominal value.
There is no way for global activation (all devices in the circuit) of mismatch simulation anymore.


hpretl:
@dwarning I don't understand the rationale why mismatch has to be enabled on a per-device basis. If I want a mismatch simulation (i.e., simulate what is really going on in silicon, compared to the simplistic and wrong situation that all elements match perfectly) then I need the mismatch active in all elements. It would be extremely cumbersome (and quite counterintuitive to have to manually enable mismatch on all elements in a large circuit).

I have also never seen such an implementation in a PDK. It is either no mismatch or mismatch on all elements. In rare case (mainly for debugging what is going on) designers want to selectively turn on mismatch on elements. In some simulators, there are functions available to provide a sensitivity analysis, which also helps in this regard.

hpretl:
@dwarning I am definitely opting for the default being `mm_ok=1`. This is what the **default** state in simulation should be, that is, conforming to reality as much as possible. Once I am in debugging/investigation mode, then I anyway tweak around to see what is going on, and turning on/off mismatch deliberately on selected instances is one way to go (another is to replace instances by "perfect" R and C from the sim lib, etc.).

dwarning:
Because there is no progress anymore applying all the contributions it makes no sense at the moment to touch the involved files. I am fearing also merge conflicts with later PR's like #476.

Issue #480, opened 2025-05-08
Error in OP annotation of npn13G2 transistors


p-fath:
### Issue

The OP annotation of the npn13g2 transistor seems to be incorrect. In addition, the Early effect is missing.

### Affected Models

-  npn13g2

### Issue Details

The saved parameters q.xq1.qnpn13g2[vbe] and q.xq1.qnpn13g2[vbc] do not match to the voltages of labeled wires, as shown in the attached picture.  

The small-signal parameters can be saved using:
save @q.xq1.qnpn13g2[gm]
.save @q.xq1.qnpn13g2[go]
.save @q.xq1.qnpn13g2[gmu]
.save @q.xq1.qnpn13g2[gpi]
.save @q.xq1.qnpn13g2[gx]
.save @q.xq1.qnpn13g2[vbe]
.save @q.xq1.qnpn13g2[vbc]
.save @q.xq1.qnpn13g2[ib]
.save @q.xq1.qnpn13g2[ic]
.save @q.xq1.qnpn13g2[cbe]
.save @q.xq1.qnpn13g2[cbc]
.save @q.xq1.qnpn13g2[cbep]
.save @q.xq1.qnpn13g2[cbcp]
...
 and can be displayed after OP analysis using the ngspice command "show".

 In addition, the go is negative, and the early effect is missing, as shown in the second picture using a DC sweep of Vce.
![Image](https://github.com/user-attachments/assets/7c39f611-16df-4a7f-959b-34da744b7355)
![Image](https://github.com/user-attachments/assets/b4258ee3-f0a7-47e2-8caa-da044b232c8e)

sergeiandreyev:
Hi @dwarning, could you please take a look?
> However, the small-signal parameters q.xq1.qnpn13g2[vbe] and q.xq1.qnpn13g2[vbc] do not match to the voltages of labeled wires, as shown in the attached picture. They seem to be modeled as the internal vbe and vbc voltages, but if this is the case they should be labeled as vbei and vbci!

dwarning:
Yes, you are correct, all shown values are operating point information's. They are internal values and someone only approximations. Beside this few of them are buggy as shown with the negative go.
[...]

PR #488, opened 2025-05-13
Self-heating of HBT devices enabled by default

p-fath:
The self-heating model was enabled in the HBT model files to increase the accuracy of the HBT DC operating points.
As shown in the plot below, the output characteristic now exactly matches the measurement results shown in 
ihp-sg13g2/libs.doc/meas/HBT/doc/html_npn13g2_VBIC/reszoom_foutput_vb.htm

Fixes a part of #480

- [x] selft switched to 1 in sg13g2_hbt_mod.lib and sg13g2_hbt_mod_mismatch.lib
- [x] Ic over Vce with constant Vbe matches measurements in ihp-sg13g2/libs.doc/meas/HBT/doc/html_npn13g2_VBIC/reszoom_foutput_vb.htm
- [x] Thermal 1GOhm resistor Rt added to sg13g2_hbt_mod.lib and sg13g2_hbt_mod_mismatch.lib

image

### DIFF
--- a/ihp-sg13g2/libs.tech/ngspice/models/sg13g2_hbt_mod.lib
+++ b/ihp-sg13g2/libs.tech/ngspice/models/sg13g2_hbt_mod.lib
@@ -42,7 +42,7 @@
 .param Nx=1 dtemp=0
 +Ny=1 le=0.96e-6 we=0.12e-6
 +El=le*1e6
-+selft=0
++selft=1
 +sw_nqs=0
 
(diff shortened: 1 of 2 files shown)


dwarning:
I want only comment that we are over the max. Ic limit for a 1x0.007x0.9 emitter stripe. The process spec said 3mA. So I would prefer to stay with the default 0 for selft and let it in the decision of the designer to switch on in case the power dissipation comes over 1mW for the Nx=1 device. Not to forget that non-convergence of larger circuits than one device can appear.

PR #501, opened 2025-05-16
DRC: add missing TV1.c/d checks

urish:
Added TV1.c and TV1.d checks to maximal deck


akrinke:
@sergeiandreyev Looks good! I just don't know if the behavior is exactly like in the commercial PDK.

akrinke:
@sergeiandreyev I'm preparing a new pull request that should be ready this week that will add these rules (and some others). They are implemented in a slightly different way to recognize TopVia1 shapes that are not enclosed at all.

akrinke:
I have added these rules in PR #538.

PR #511, opened 2025-05-23
Changed ptap_holes calculation

### LINKED ISSUE #498: sealring produces short-circuit in LVS
i have a strange problem:
a chip that seems ok without sealring in the LVS is short-circtuit in LVS if it has a sealring. but if tracing the nets in KLaout it looks ok with and without sealring. 

i was thinking that the difference of LVS and net-traceing is LVS looks also at the fill and tracing not, but if i delete all the fills it changes nothing.

but see Yourself 

[gds_and_spice.zip](https://github.com/user-attachments/files/20227533/gds_and_spice.zip)

there is still an LVS problem without sealring that comes from two parallel cmim's its not really a problem, its an error of the LVS handling of parallel cmim (at least i think this).

this is VSS without sealring

Image

and this is VSS including sealring

Image

d-m-bailey:
The holes function returns the inside of donut shapes. The seal ring is a ptap that surrounds the entire chip. When the sealring is included, the entire chip becomes `ptap_holes`. This change ignores the sealring ptap when calculating ptap_holes. For designs without the sealring, there is no change. 
Warning: Use of the holes function in LVS can cause unexpected side-effects when macros have enclosing tap rings.

Fixes #498

- [?] Tests pass <- This change is part of a larger set of changes that I'm working on for the TO_Apr2025 shuttle tests. The larger set of changes is still being tested, but this specific change has been verified to solve the issue above.
- [X] Appropriate changes to README are included in PR <- No changes to README

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/lvs/rule_decks/general_derivations.lvs
+++ b/ihp-sg13g2/libs.tech/klayout/tech/lvs/rule_decks/general_derivations.lvs
@@ -88,7 +88,7 @@ ptap1_mk = substrate_drw.and(pwell).interacting(ptap1_lbl)
 # n & p taps (short connections)
 ntap = nactiv.and(nwell_drw).not(ntap1_mk).not(recog_diode).not(gatpoly)
 ptap = pactiv.and(pwell).not(ptap1_mk).not(recog_diode).not(gatpoly)
-ptap_holes = ptap.holes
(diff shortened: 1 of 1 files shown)


KrzysztofHerman:
@atorkmabrains @FaragElsayed2 could you please confirm the validity of this PR ?

FaragElsayed2:
@KrzysztofHerman Confirmed

Issue #418, opened 2025-03-30
Device Present in Netlist but not showing during LVS


rupokdas147:
Capacitor **CDC1** is present in netlist but not showing in the reference during LVS check.

![Image](https://github.com/user-attachments/assets/6436815a-d9be-4b23-b2b3-df8e8b621ab5)

![Image](https://github.com/user-attachments/assets/d21c543f-3b1a-42d4-bc8c-79864286a61a)

d-m-bailey:
@rupokdas147 In the source side, `CDC1` and `CDC3` have been combined to `DC3` (check the sizes).

On the layout side, the devices `$6` and `$2` have not been combined. Can you verify that `$6` and `$2` are parallel (connected the same) in the layout?

There may be a problem with the reduction rules. It appears that `w`, `l`, and `A` are all summed, while `P` remains the same.

rupokdas147:
@d-m-bailey yes they are parallel in the layout side. This is the netlist provided by Klayout.

![Image](https://github.com/user-attachments/assets/0daf93f9-f527-4f66-aa3f-584e5484d579)

sergeiandreyev:
Hi @atorkmabrains, @FaragElsayed2, could you please take a look?

KrzysztofHerman:
Hi @rupokdas147, PRs #800, #828, and #862 have been merged. Do these changes resolve the issue? If yes, please consider closing this issue. Thanks!

PR #513, opened 2025-05-23
Corrected issues with rfcmim extraction.

### LINKED ISSUE #459: LVS Problem of two in series connected capacitor
Hi
I have two in series connected RFCMIMs and KLayout throws me this error

Image

Why?
Layout:

Image

Schematic:

Image

BW, Ghaith.

### LINKED ISSUE #512: rfcmim LVS problems.
1. A design with multiple `rfcmim` capactors gives warnings during extraction
```
2025-05-23 10:12:23 +0900: Memory Usage () : Extracting rfcmim device
2025-05-23 10:12:23 +0900: Memory Usage () : No. of ports exist for rfcmim topmetal is 9, should be 1
```
2. parallel or series devices in the source netlist cause LVS to crash.
```
2025-05-23 10:12:24 +0900: Memory Usage () : Starting SG13G2 LVS Simplification
ERROR: In /Users/home/IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/tech/lvs/sg13g2.lvs: Internal error: Terminal still connected after removing device in device combination: name=RC3, circuit=FMD_QNC_04_160GHZ_LNA, terminal=mim_sub in Netlist::combine_devices
ERROR: RuntimeError: Internal error: Terminal still connected after removing device in device combination: name=RC3, circuit=FMD_QNC_04_160GHZ_LNA, terminal=mim_sub in Netlist::combine_devices in Executable::execute
  /Users/home/IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/tech/lvs/sg13g2.lvs:362:in `combine_devices'
  /Users/home/IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/tech/lvs/sg13g2.lvs:362:in `execute'
  :/built-in-macros/lvs_interpreters.lym:31:in `instance_eval'
  :/built-in-macros/lvs_interpreters.lym:31:in `execute'
```
3. device extracted as `rfcmim` but spice model is `cap_rfcmim`.

test case: 

[test_rfcmim.tgz](https://github.com/user-attachments/files/20402631/test_rfcmim.tgz)
```
tar xzf test_rfcmim.tgz
cd test_rfcmim
export PDK_ROOT=<path_to_pdk>
./run_lvs
```

Pull request to fix these issues will be submitted shortly.

d-m-bailey:
Changed the extracted device model from rfcmim to cap_rfcmim to match spice model. 
Changed rfmim_sub calculation.
Added matching mim_sub connection to parallel cap_rfcmim check. 
Use custom combiner for both layout and source reduction.

Fixes #512 Also may fix #459 and #418

- [O] Tests pass (test case in #512 completes without giving warnings, without crashing, and with models that match spice.)
- [X] Appropriate changes to README are included in PR <- no README changes.


KrzysztofHerman:
@sergeiandreyev please take care

sergeiandreyev:
Hi @d-m-bailey, could you please roll-back this part of the PR:
> Changed the extracted device model from rfcmim to cap_rfcmim to match spice model.

As you can see, now the LVS regression is failing.. we have different device names for the `rfcmim` capacitor based on which netlist type you're generating, for simulation (SPICE) or LVS (CDL).
The simulation netlist should contain `cap_rfcmim` and the netlist for LVS `rfcmim`, and, btw, this has to be updated for xschem to generate the CDL netlist in correct way. It is not there yet

d-m-bailey:
@sergeiandreyev 

> As you can see, now the LVS regression is failing.. we have different device names for the rfcmim capacitor based on which netlist type you're generating, for simulation (SPICE) or LVS (CDL).

I see. 

With different model names, I see potential problems. 

Currently, the klayout rules extract as `rfcmim` for LVS. I believe the current magic tech file does not support `rfcmim`. I think my PR adds the `cap_rfcmim` devices to magic. Magic is also used for parasitic extraction - meaning that the extracted netlist will be simulated in spice. So magic would need 2 extraction rules (which is possible with variants), one to extract parasitic devices with `cap_rfcmim` and one to extract with `rfcmim` for LVS.

What is the reason for different model names between simulation and LVS?

sergeiandreyev:
> What is the reason for different model names between simulation and LVS?

first of all, this is done with alignment to commercial PDK, while for the actual reason - let me double check on this internally..
from my point of view, we have somewhat confusing (and not clear) approach here - some devices have the same names between different netlist types, other have different, etc.

d-m-bailey:
@sergeiandreyev 

> > What is the reason for different model names between simulation and LVS?
> 
> 
> first of all, this is done with alignment to commercial PDK, 
I thought that might be the case.

Recently, bulk nodes have been added to the poly resistors for simulation. Are there plans to add those to LVS also or do we continue the separate LVS and simulation netlist approach.

d-m-bailey:
@RTimothyEdwards I added `rfcmim` rules (mimicking `cap_cmim`) to the magic and netgen tech files. Can you do a quick sanity check?

d-m-bailey:
@FaragElsayed2 Can you check the changes to the klayout rule deck? This PR modifies the extraction of the substrate terminal for `rfcmim` and also fixes a problem where the combination routine was different between the layout and source.

FaragElsayed2:
@OmarMahXII Could you please check those LVS changes matched with our updates?

d-m-bailey:
#663 appears to include all of the klayout changes.

The magic tech file changes are not included in #663

@RTimothyEdwards the LVS model `rfcmim` is different than the simulation model `cap_rfcmim`. Rather than having 2 versions in the magic tech file, I was thinking of having magic always extract as `cap_rfcmim` (for compatibility with parasitic extraction) and establishing a correspondence between `rfcmim` and `cap_rfcmim` in the netgen setup file. What do you think?

RTimothyEdwards:
@d-m-bailey :  If `rfcmim` is not a simulatable device, why would you extract it?
Do not make these changes to the magic tech file.   If this should be handled at all, it should be done with a style variant, not the way you are doing it.
I am developing the magic tech file under contract with IHP.  If you make changes to it, you are interfering with my work.

RTimothyEdwards (commented):
Changes to the magic tech file are not approved;  I am in the middle of developing the tech file and will need to make changes to magic and netgen to support the misguided use of CDL netlists for LVS.

d-m-bailey:
@RTimothyEdwards 
Thanks for the feedback.

> If rfcmim is not a simulatable device, why would you extract it?

Excellent point! Please see the discussion [here](https://github.com/IHP-GmbH/IHP-Open-PDK/pull/513#issuecomment-3012912092). It's probably for the same reason that you are developing a tech file for  "the misguided use of CDL netlists for LVS". 😉

It seems to me that IHP wants to use the same netlist for klayout LVS, magic/netgen LVS, and proprietary tools. However, since klayout (and proprietary tools?) does not support the simulation netlist, xschem has to be configured to generate 2 styles of netlist. It already does this to some extent (netlisted parameters are different based on the lvs flag), but the modifications needed will be more extreme (ie, 2 different spice prefixes and formats for every primitive model).

In other processes, either the klayout source netlist is created by modifying the netgen source netlist or vice versa. I don't see the merit of forcing both tools to accept the same format. 

We're not under contract (yet!), but we're using previous open-source shuttle data to run LVS (magic/netgen and klayout), soft connection checks, and CVC and reporting any problems (and possible solutions) that we encounter. 

I don't want to interfere with your work, so going forward, I plan to _raise_ any pdk related magic/netgen issues that I come across, but I _will not_ submit pull requests. Is this acceptable?

Just to be clear, pull requests for magic and netgen changes are ok, but not IHP magic tech file changes, right?

CC: @H-Ojiro

sergeiandreyev:
Hi all,
 regarding this
 > xschem has to be configured to generate 2 styles of netlist

indeed, open source schematic editors which are officially supported by IHP OpenPDK (xschem and Qucs-S) should have the capability to generate two netlist types (and if this is not supported, it should be implemented):
1. netlist for simulation
2. netlist for LVS (`CDL`)

The first one should not be used for LVS, the second one should not be used for simulation. And the open source LVS tools/engines should support the second type - this way we will have alignment between open source and commercial flows.
We could have a separate meeting to discuss this and try to resolve all outstanding questions.

d-m-bailey:
@sergeiandreyev Thanks for the response. That pretty much confirms what I suspected. 

What about parasitic extraction? You'd want the layout extractor to be able to extract a simulatable version in addition to an LVS version, right?

magic does have the ability to extract different styles of netlists based on variants, but in other processes, we've used the simulatable version for LVS too.

I'm open to a meeting, but since I don't own the rule decks, I'd just be there to give my opinions.

RTimothyEdwards:
@sergeiandreyev :  I have made my views on this known in the past, but in my opinion, as soon as you have to have a _different_ netlist for simulation than for LVS, then LVS is no longer a proper validation.  And there is no reason why a netlist extracted for simulation should not be able to be used for LVS, and no reason why a netlist extracted for LVS should not be able to be used for simulation.

The CDL netlist format has the additional major issue that, as far as I know, there is no standard for the format, it is not clearly an open standard, and the syntax is possibly different for different commercial tools.  It is not even clear to me that an open source tool can legally write a CDL format file.

sergeiandreyev:
Hi @RTimothyEdwards, @d-m-bailey, I propose to have a meeting somewhere on W38

d-m-bailey:
@sergeiandreyev W38 is September 14th - September 20th, right? Let me know what times are good for you.

RTimothyEdwards:
Should work for me as well.

sergeiandreyev:
@d-m-bailey , I have to get feedback from IHP people to align on possible time slots that week, I'll come back with the proposal

PR #520, opened 2025-05-28
Added bulk node to resistor models rhigh + rppd + rsil

p-fath:
For analog designs, it would be very advantageous to have an explicit bulk node in the symbols of the resistors to be able to compensate for their parasitic capacitance. 

- [x]  symbol of rppd, rsil and rhigh extended with a dedicated bulk pin
- [x] models of rppd, rsil and rhigh updated for ngspice
- [x] models of rppd, rsil and rhigh updated for xyce, **as parasitic capacitances are not modeled in the actual implementation with xyce, the additional node is not used in xyce simulations**


dwarning:
@p-fath : I try to understand your intention. At the moment we have parasitic capacitances from r3_cmc model to ground node. This is fixed in the used subckt's in ngspice and xyce lib. The subckt has only the two terminals for the resistor. Do you want that the 3'rd node from the r3_cmc instance (now on 0) should be in the subckt list e.g. _.subckt rsil 1 2 b_ ?
The difference would be that in case the resistor goes over an well etc. you can connect the r3 parasitic capacitances to this node. But from rf point of view is it not the same because any well bias is rf ground?
Sorry for asking this - only to understand.

dwarning (commented):
Dependent from the answer from TO and the usefulness we should extend the PR also to ngspice lib and merge it into dev branch.

p-fath:
Many thanks for checking out this PR!
Yes, we would like to be flexible with setting the bulk node of the resistor; it should not be fixed to GND (0).
As an explanation, I want to use this example circuit (sizing not correct!):
image
Here, the parasitic capacitance of R1 is compensated (resulting in a higher BW) as the low-ohmic voltage follower drives the bulk node of R1 to VOUT and VOUT=VX. As a result, the substrate (well) of R1 is on the same voltage level as VX -> deltaV=0V, resulting in deltaQ=C*deltaV=0C. From an RF perspective, this also holds within certain limits. If the bulk node of R1 were driven with a fixed bias voltage, it would be AC gnd or RF gnd as you already explained.

dwarning:
This sounds reasonable.
@KrzysztofHerman will check compatibility to commercial PDK and come back here.

KrzysztofHerman:
> This sounds reasonable. @KrzysztofHerman will check compatibility to commercial PDK and come back here.

we have to discuss it internally. I am pushing to integrate it. As an alternative we could also have symbols which do not have an explicit 3-rd terminal but are by default connected to the `sub!` net (substrate/bulk), what could be changed by the user providing the feature to connect to any network. An example of such symbol goes here:

image

d-m-bailey:
Will this require changes to the klayout/magic extraction rule files to handle 3 transistor resistors?

KrzysztofHerman:
@p-fath I have pushed another PR #563 where I have changed the symbols. Please take a look at the "known issues"

KrzysztofHerman:
> Will this require changes to the klayout/magic extraction rule files to handle 3 transistor resistors?

@d-m-bailey yes it will. Right now the `klayout` LVS extraction extracts 2 terminal device.

PR #526, opened 2025-05-29
ihp-sg13g2: libs.tech: klayout: Update sg13g2.map File

### LINKED ISSUE #345: KLayout: wrong LEF/DEF mapping for boundary
It seems that the KLayout LEF/DEF mapping uses the wrong layer:datatype pair for `DIEAREA`:

https://github.com/IHP-GmbH/IHP-Open-PDK/blob/4c6508d03a3078b21c737d04fae5dccec9aa590f/ihp-sg13g2/libs.tech/klayout/tech/sg13g2.map#L154

235:4 is the layer:datatype for `prBndry` from sky130: https://skywater-pdk.readthedocs.io/en/main/rules/layers.html

Taking a look at the "SG13G2 Layout Rules" it seems that 189:0 is the correct layer:datatype:

Image

But if we look at the `sg13g2_stdcell` cells we can see that 189:4 is used exclusively for the boundary.

The  KLayout layer property files designate 189:4 as prBoundary.boundary:

https://github.com/IHP-GmbH/IHP-Open-PDK/blob/4c6508d03a3078b21c737d04fae5dccec9aa590f/ihp-sg13g2/libs.tech/klayout/tech/sg13g2.lyp#L5664

So it seems that 189:4 is the correct layer:datatype pair for `DIEAREA`.

---

In summary:

1. The KLayout LEF/DEF mapping uses the wrong layer:datatype for the boundary
2. The documentation is missing 189:1 and 189:4 and should probably state that 189:4 should be used as boundary
3. I'm not sure what the `COMP` statement in the LEF/DEF mapping is used for, maybe someone can clarify that?

Thanks!

PS: It seems the LEF/DEF mapping was initially copied from the OpenROAD-flow-scripts: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/platforms/ihp-sg13g2/sg13g2.map
So it should probably either be removed or updated on there.

### LINKED ISSUE #442: sram macros adding shapes to 235/4
The current sram macros have boxes at layer `prBoundary.boundary`.

(purple boxes)

Image

After merging the sram macro gds into a chip with OpenROAD, these boxes appear on 235/4

(sram macro is rotated by 180 degree!)

Image

### LINKED ISSUE #518: Running into Pin.f violations
OpenROAD-flow-scripts is able to generate macro blocks for Verilog modules. Those blocks have pin shapes for routing signals to in the parent design hierarchy.

Image

When exporting a design to GDS, those pin shapes will be added to data layer 2 (`MetaX.pin`). According to the Layout Rules, pin areas ust be fully covered by drawing

```
Circuit designers should use only drawing purpose 0 (data types) for layouts. Only exception is pin purpose
2 for symbolic pins. Data type 2 (purpose pin) is used for symbolic connectivity information. Pin areas must
be fully covered by drawing. These rules are tested because pin areas are not used for mask generation
and a potential issue due to false postive LVS matchs.
```

Unfortunately, that is not possible with OpenROAD and designs with self-generated blocks will have `Pin.f` violations as a short part of the track is not covered by `MetalX.drawing`.

Image

**Questions:**

1. Why is data layer 2 not used to generate masks and what's the purpose of data layer 2 then?
2. Can we export data layer 2 (``pin``) to data layer 0 (``drawing``)?

dnltz:
* Remove all unused NAME layer names that were commented out.
* Align all layers properly with tabs to have them at the indentation.

I will update the `sg13g2.map` file in OpenROAD-Flow-Scripts afterwards.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/sg13g2.map
+++ b/ihp-sg13g2/libs.tech/klayout/tech/sg13g2.map
@@ -1,133 +1,109 @@
-#************************************************************************#
-#************************************************************************#
-# File		: sg13g2.map
-# DATE		: September 28, 2022
+#************************************************************************
+#************************************************************************
+# File: sg13g2.map
+# DATE: September 28, 2022
+#************************************************************************
+#************************************************************************
+#
+# Copyright 2023 IHP PDK Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
 #*************************************************************************
-#************************************************************************* 
-#                                                                         
-# Copyright 2023 IHP PDK Authors                                         
-#                                                                       
-# Licensed under the Apache License, Version 2.0 (the "License");       
-# you may not use this file except in compliance with the License.      
-# You may obtain a copy of the License at                                     
-#                                                                             
-#     https://www.apache.org/licenses/LICENSE-2.0                             
-#                                                                             
-# Unless required by applicable law or agreed to in writing, software         
-# distributed under the License is distributed on an "AS IS" BASIS,           
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.    
-# See the License for the specific language governing permissions and         
-# limitations under the License.                                              
-#                                                                             
-#************************************************************************* 
 #-------------------------------------------------------------------------
 # EDI Stream layer mapping table for SG13
-# Version: 1.0 
+# Version: 1.0
 # Date: 22 Aug 2023
 # Use only in combination with valid GDSII data of all used blocks!
 #-------------------------------------------------------------------------
-#EDI Layer Name	  EDI Layer Type   GDS Layer Number   GDS Layer Type
-#==============	  ==============   ================   ==============
+#EDI Layer Name  EDI Layer Type               GDS Layer Number   GDS Layer Type
+#==============  ==============               ================   ==============
 
-Metal1		NET,SPNET,PIN,LEFPIN,VIA	8		0
-Metal1		PIN,LEFPIN					8		2
-Metal1		FILL						8		22
-Metal1		LEFOBS						8		4
+Metal1           NET,SPNET,PIN,LEFPIN,VIA     8                  0
+Metal1           PIN,LEFPIN                   8                  2
+Metal1           FILL                         8                  22
+Metal1           LEFOBS                       8                  4
 
-#NAME		Metal1/NET		20		0
-#NAME		Metal1/SPNET	20		0
-NAME		Metal1/PIN		8		25
-#NAME		Metal1/LEFPIN	20		0
+NAME             Metal1/PIN                   8                  25
 
-Via1		PIN		19		0
-Via1		LEFPIN	19		0
-Via1		VIA		19		0
+Via1             PIN                          19                 0
+Via1             LEFPIN                       19                 0
+Via1             VIA                          19                 0
 
-Metal2		NET,SPNET,PIN,LEFPIN,VIA	10		0
-Metal2		PIN,LEFPIN					10		2
-Metal2		FILL						10		22
-Metal2		LEFOBS						10		4
+Metal2           NET,SPNET,PIN,LEFPIN,VIA     10                 0
+Metal2           PIN,LEFPIN                   10                 2
+Metal2           FILL                         10                 22
+Metal2           LEFOBS                       10                 4
 
-#NAME		Metal2/NET		21		0
-#NAME		Metal2/SPNET	21		0
-NAME		Metal2/PIN		10		25
-#NAME		Metal2/LEFPIN	21		0
+NAME             Metal2/PIN                   10                 25
 
-Via2		PIN		29		0
-Via2		LEFPIN	29		0
-Via2		VIA		29		0
+Via2             PIN                          29                 0
+Via2             LEFPIN                       29                 0
+Via2             VIA                          29                 0
 
+Metal3           NET,SPNET,PIN,LEFPIN,VIA     30                 0
+Metal3           PIN,LEFPIN                   30                 2
+Metal3           FILL                         30                 22
+Metal3           LEFOBS                       30                 4
 
-Metal3		NET,SPNET,PIN,LEFPIN,VIA	30		0
-Metal3		PIN,LEFPIN					30		2
-Metal3		FILL						30		22
-Metal3		LEFOBS						30		4
+NAME             Metal3/PIN                   30                 25
 
-#NAME		Metal3/NET		22		0
-#NAME		Metal3/SPNET	22		0
-NAME		Metal3/PIN		30		25
-#NAME		Metal3/LEFPIN	22		0
+Via3             PIN                          49                 0
+Via3             LEFPIN                       49                 0
+Via3             VIA                          49                 0
 
-Via3		PIN		49		0
-Via3		LEFPIN	49		0
-Via3		VIA		49		0
+Metal4           NET,SPNET,PIN,LEFPIN,VIA     50                 0
+Metal4           PIN,LEFPIN                   50                 2
+Metal4           FILL                         50                 22
+Metal4           LEFOBS                       50                 4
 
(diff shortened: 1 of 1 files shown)


dnltz:
@volkermuehlhaus - for some reason I can't see your comment here. That's what I received as mail
```
This is not a good idea. Pin purpose is required. If you have an issue with pin shapes, something else is wrong.
```

- Can you tell me what's the purpose of Pin?
- Why is this data layer ignored during mask creation?

I described the issue here: https://github.com/IHP-GmbH/IHP-Open-PDK/issues/518

VolkerMuehlhaus:
> for some reason I can't see your comment here.

Please ignore my comment! I misunderstood something and already deleted my comment here.

smunaut:
Why would you remove the layer `2` pin ?!?!
Yes it's not a mask layer used for generating manufacturing masks, but it's still a useful information layer ! That can be used on import to define which area are meant to be connected to for instance ...

If you have a problem with OpenROAD-Flow-Scripts creating `Pin.f`, are you sue the map file in there is up to date and in sync with the current one from the PDK ?

smunaut:
Also `DIEAREA` is also useful, it's just using the wrong layer ATM, see #345  for the proper fix which is using the proper `189` instead of `235` there.

```diff
-COMP           ALL             235             0
+COMP           ALL             189             0
 
-DIEAREA                ALL             235             4
+DIEAREA                ALL             189             4
```

mole99:
It seems this PR is a response to: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/3191
But I'm not sure what you're trying to achieve here?

With the fix Sylvain mentioned, I could do a hierarchical implementation on [Greyhound](https://github.com/mole99/greyhound-ihp) just fine.

dnltz:
> Also `DIEAREA` is also useful, it's just using the wrong layer ATM, see #345 for the proper fix which is using the proper `189` instead of `235` there.
> 
> ```diff
> -COMP           ALL             235             0
> +COMP           ALL             189             0
>  
> -DIEAREA                ALL             235             4
> +DIEAREA                ALL             189             4
> ```

I can change that.

dnltz:
> It seems this PR is a response to: [The-OpenROAD-Project/OpenROAD-flow-scripts#3191](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/3191) But I'm not sure what you're trying to achieve here?
> 
> With the fix Sylvain mentioned, I could do a hierarchical implementation on [Greyhound](https://github.com/mole99/greyhound-ihp) just fine.

Can you please share you branch? The current `.map` is not working with hierarchical blocks.

VolkerMuehlhaus:
@dnltz Let me come back to your questions above, maybe there is a simple misunderstanding of the pin shape. 
This is used to identify device terminals, but it always sits on top of the drawn metal (purpose drawing). 

I guess you might have drawn the pin shape **instead of** shape with purpose drawing, that would be wrong.

smunaut:
@dnltz The .map in ORFS is _wrong_.  You need to copy the .map that is in this repository and overwrite the ORFS one.

dnltz:
> @dnltz The .map in ORFS is _wrong_. You need to copy the .map that is in this repository and overwrite the ORFS one.

That's exactly what I did...

dnltz:
> @dnltz Let me come back to your questions above, maybe there is a simple misunderstanding of the pin shape. This is used to identify device terminals, but it always sits on top of the drawn metal (purpose drawing).
> 
> I guess you might have drawn the pin shape **instead of** shape with purpose drawing, that would be wrong.

Just for clarification: OpenROAD drew those pin shapes when generating a block macro.

smunaut:
The `.map` in this repo will generate both `drawing` and `pin` GDS layer types for any `PIN` / `LEFPIN` from the LEF/DEF input. So it can't create anything with `Pin.f` violations ...

It works that way because what GDS data type `pin` and what the LEF/DEF `PIN` / `LEFPIN` means is a bit different. In the GDS it's purely informative, marking an area as being a contact zone but doesn't generate any "metal" on the masks. In the LEF/DEF world, the `PIN` area are expected to create metal where they are. 

(and note that this is the same in SKY130 and GF180, this is not an IHP thing).

mole99:
> > Also `DIEAREA` is also useful, it's just using the wrong layer ATM, see #345 for the proper fix which is using the proper `189` instead of `235` there.
> > ```diff
> > -COMP           ALL             235             0
> > +COMP           ALL             189             0
> >  
> > -DIEAREA                ALL             235             4
> > +DIEAREA                ALL             189             4
> > ```
> 
> I can change that.

@dnltz Please don't. I'm in the process of preparing a PR for LibreLane that includes this change.

dnltz:
> The `.map` in this repo will generate both `drawing` and `pin` GDS layer types for any `PIN` / `LEFPIN` from the LEF/DEF input. So it can't create anything with `Pin.f` violations ...
> 
> It works that way because what GDS data type `pin` and what the LEF/DEF `PIN` / `LEFPIN` means is a bit different. In the GDS it's purely informative, marking an area as being a contact zone but doesn't generate any "metal" on the masks. In the LEF/DEF world, the `PIN` area are expected to create metal where they are.
> 
> (and note that this is the same in SKY130 and GF180, this is not an IHP thing).

Thanks for the explanation. I didn't know it's supposed to generate shapes in both layer types. I will take a look in OpenROAD/Klayout why the `drawing` is missing.

dnltz:
> > > Also `DIEAREA` is also useful, it's just using the wrong layer ATM, see #345 for the proper fix which is using the proper `189` instead of `235` there.
> > > ```diff
> > > -COMP           ALL             235             0
> > > +COMP           ALL             189             0
> > >  
> > > -DIEAREA                ALL             235             4
> > > +DIEAREA                ALL             189             4
> > > ```
> > 
> > 
> > I can change that.
> 
> @dnltz Please don't. I'm in the process of preparing a PR for LibreLane that includes this change.

Can you please fix that after this pull-request? This issue has nothing to do with LibreLane ...

mole99:
As I said, it is part of my LibreLane PR. Otherwise the XOR between KLayout and magic GDS would fail.

dnltz:
this pull-request is now only fixing indentation and removing all lines that were commented out. @KrzysztofHerman  @sergeiandreyev can you take a look?

dnltz:
Let's wait for #540 first

stefanottili:
a) PIN is the only map entry that is mapped twice, to two different layer:datatype
One for geometry, since net geometry might end in the middle of the pin.
The other to be able recognize pin geometry when doing abstract generation from gds.

b) I've never seen any technology where NET,SPNET,VIA,PIN wouldn't be mapped to the same layer:datatype
Either write it as comma separated list or use ALL, both can be written on a single line.

c) NAME ALL
This shall never be in a map file.
It creates TEXT for every single def COMPONENT, including tap/filler cells.
Not only not needed for any downstream tool, but very bad for tool performance, needlessly filling hash tables.
As bad in gds as it is in oasis, since all the instance names are unique.

d) COMP ALL
This shall never be in a map file.
It creates a bbox rectangle for every single def COMPONENT, including tap/filler cells.
Needlessly bloats gds with geometry that no tool downstream has ever used.
Not as bad in oasis as in gds, since oasis has repetition 10/11, but still useless information.

dnltz:
> a) PIN is the only map entry that is mapped twice, to two different layer:datatype One for geometry, since net geometry might end in the middle of the pin. The other to be able recognize pin geometry when doing abstract generation from gds.

That issue got solved by syncing this file from here to ORFS.
> 
> b) I've never seen any technology where NET,SPNET,VIA,PIN wouldn't be mapped to the same layer:datatype Either write it as comma separated list or use ALL, both can be written on a single line.
> 
> c) NAME ALL This shall never be in a map file. It creates TEXT for every single def COMPONENT, including tap/filler cells. Not only not needed for any downstream tool, but very bad for tool performance, needlessly filling hash tables. As bad in gds as it is in oasis, since all the instance names are unique.

I don't see `NAME ALL` in this file.
> 
> d) COMP ALL This shall never be in a map file. It creates a bbox rectangle for every single def COMPONENT, including tap/filler cells. Needlessly bloats gds with geometry that no tool downstream has ever used. Not as bad in oasis as in gds, since oasis has repetition 10/11, but still useless information.

I think you should open an issue requesting to remove this line. This PR has no functional changes anymore since my original issues got fixed. It just makes the `.map` file better to read and removes commented out lines.

stefanottili:
see https://github.com/IHP-GmbH/IHP-Open-PDK/issues/548

NAME COMP is used in OpenROAD-flow-scripts/flow/platforms/ihp-sg13g2/sg13g2.map
I'm assuming that eventually they will sync with IHP-Open-PDK in the future.

maliberty:
If you need the PDK updated in ORFS let me know.  We don't tend to care about the GDS much in terms of OR testing.

sergeiandreyev:
ok, I will merge this one - then the discussion should continue in #548

sergeiandreyev (approved):

PR #539, opened 2025-06-12
SRAM 64x32

b10346:
Added SRAM block:
- 64x32
- Dual Port
- No BM
- No BIST

**IMPORTANT: Input ports `A_DLY` / `B_DLY` must be tied to “1” always.** – This requirement is valid for all IHP SRAM blocks.


sergeiandreyev:
Hi @dnltz, could you please check this PR?

Issue #431, opened 2025-04-04
OpenLane support


jadoro:
Hello,

I was trying to synthesisze a small test deisgn for `ihp-sg13g2` using OpenLane and found that currently (68eebaf), no `openlane/config.tcl` is provided in `libs.tech`.
Digging a bit, I found [this IHP press release](https://www.ihp-microelectronics.com/services/research-and-prototyping-service/fast-design-enablement/open-source-pdk), stating
> Within the project there will be two established open source digital design flows that can be used with the IHP Open Source PDK. While the OpenROAD flow is already supported, the setup for the OpenLane flow will be provided within 2024.

I just wanted to kindly ask whether there is any update on this statement, especially given the [unclear future of efabless](https://web.archive.org/web/20250324223745/https://efabless.com/notice), who seemed to be a major driving force behind OpenLane.

Thanks a lot for your efforts!

KrzysztofHerman:
@jadoro there is ongoing WiP to have support for openlane you can try it out using [this](https://github.com/mole99/IHP-Open-PDK/tree/openlane) fork & branch. Maybe @mole99 could give you some more details.

jadoro:
@KrzysztofHerman That looks great, thanks a lot for the hint! I'll give it a try.

KrzysztofHerman:
@jadoro give us some feedback about the performance and issues if you find any.

mole99:
I have a few examples ready to run: https://github.com/mole99/ihp-openlane-examples
(Note, the configs may need to be adjusted, i.e. the die area for ttihp0p2-one-sprite-pony)

For simple macros it already works fine, but there are still a few issues that need to be sorted out.
I've been designing a chip for the April tapeout using OpenLane, and there were a few issues for which fixes still need to be properly integrated. This is going to happen during the next few months.

jadoro:
Thanks a lot, this looks like a really nice example to get started with!

We're currently using OpenLane 1, which seems to expect slightly different variables in `config.tcl` (e.g. `CELLS_LEF` instead of `CELL_LEFS`, `LIB_SYNTH` being defined, …). I will probably try to see if I can achieve OL1 compatibility in reasonable time, otherwise me might migrate our flow.

mole99:
You're welcome! Yes, many of the variables have been renamed or deprecated since OpenLane 2.

[...]

PR #387, opened 2025-03-07
IO cells update

b10346:
IO cells library update:
- added metal1 contacts to substrate and n-wells to fix latch up DRC errors,
- modified slightly Active and Metal1 layers to provide contacts implementation, 
- added ptap1 devices to layouts: "substrate drawing" layer  and "sub!" label in "TEXT drawing" layer were added on the area where "pSD drawing" and "Active drawing" and not "NWELL" drawing are presented,
- added ptap1 device in schematic to connect vss to substrate, iovss to substrate through resistor: CDL netlist is updated,
- I/O library cell names were updated to avoid potential collisions with another opes source projects,
- Parameter FOREIGN was changed in LEF model of corner cell,
- New file with a list of I/O library cells and sub-cells is placed in doc folder,
- DRC and LVS checks were done with internal IHP tools.
- **[IMPORTANT] Re-running filler insertion for already place-and-routed designs is necessary due to Metal1 changes.**

PR #540, opened 2025-06-12
Initial LibreLane Setup

mole99:
LibreLane is the successor to OpenLane 2, see https://github.com/librelane/librelane for more information.

This is the initial setup for LibreLane so that we all have a common base to work from.
It consists of the combined work previously done for OpenLane 2 by the Tiny Tapeout team (thanks!) and my work at Efabless as well for [Greyhound](https://github.com/mole99/greyhound-ihp/).

The setup can be found in the `librelane` directory under `libs.tech`.
This is still early work and should be used with caution.

What needs further work in follow-up PRs:

- RC-values
- Clock buffer selection (https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/discussions/2887)
- Better antenna insertion (requires newer OpenROAD)
- Padring generation

~~There is one "hack" in this PR that overcompensates Metal2 min area in order to prevent a DRC violation for the same.
This hack can be removed once https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/issues/2574 is fixed.~~

## Testing

To try this PR, the [`ihp`](https://github.com/librelane/librelane/tree/ihp) branch of LibreLane must be used.
The `ihp` branch will be soon merged into `dev` and `dev` will be released as LibreLane 3.0.

Follow the instructions in the [docs](https://librelane.readthedocs.io/en/latest/) to set up LibreLane with Nix.

To verify a working setup, run the smoke test:

```
PDK_ROOT=/path/to/IHP-Open-PDK PDK=ihp-sg13g2 librelane --manual-pdk --smoke-test
```

Supersedes #431

@smunaut @atorkmabrains @dnltz


mole99:
Hi @FaragElsayed2, I saw that you implemented the KLayout LVS scripts.
Can I ask you what the `$[...]` in the reference `.cdl` netlists mean?

For example:

```
C0 PLUS1 MINUS1 $[cap_cmim] 74.620f w=6.99u l=6.99u m=1
```

It seems this syntax allows to annotate the actual value, e.g. capacitance, of the device.

image

But it appears that the changes to the LVS reader in this PR break this behavior. Do you perhaps have any suggestions?

FaragElsayed2:
Hi @mole99,

These CDL files were originally provided by the IHP team and are typically generated by commercial tools. They serve as golden references for LVS validation. The syntax you're referring to (e.g., $[cap_cmim]) is not standard in open-source tools like Xschem and is also not recognized by the default KLayout CDL reader.

To handle such cases, I implemented a custom parser that can process both standard CDL syntax and these extended formats. This extended syntax likely originates from how the device CDFs are defined internally in the PDK. As part of the parser, I included a line to clean up such patterns before processing:

```ruby
line = line.delete('[]$\\/')
```

However, based on this [discussion](https://www.klayout.de/forum/discussion/comment/11385/#Comment_11385), it seems that the cleanup line was too generic and introduced issues in other test cases.

I believe the best approach is to handle this more selectively — applying the cleanup only in specific scenarios, such as the case you mentioned for mim. This way, we preserve compatibility with the extended syntax without unintentionally affecting other parts of the netlist.

CC @atorkmabrains

mole99:
Hi @FaragElsayed2,

Thank you for your detailed explanation! That really helped me to understand why these edits were made in the first place.

For this PR, I've now added a more targeted regex that removes the `$[...]` around the device type.
This fixes the failed tests and doesn't cause any more regressions.

If there are some issues with this in the future, we may need to create an even more targeted version, but for now this seems fine.

mole99:
Hi @sergeiandreyev and @KrzysztofHerman, this PR is now ready.

Could you rerun the failing tests? This is due to #542.

ihp-sg13g2/libs.tech/librelane/config.tcl
```diff
@@ -0,0 +1,162 @@
...
+set ::env(CELL_CDLS) "$::env(PDK_ROOT)/$::env(PDK)/libs.ref/$::env(STD_CELL_LIBRARY)/cdl/$::env(STD_CELL_LIBRARY).cdl"
+
+# GPIO Pads
+set ::env(GPIO_PADS_LEF) "\
+	$::env(PDK_ROOT)/$::env(PDK)/libs.ref/sg13g2_io/lef/sg13g2_io.lef\
+	$::env(PDK_ROOT)/$::env(PDK)/libs.ref/sg13g2_io/lef/sg13g2_io_notracks.lef\
+"
+
+set ::env(GPIO_PADS_VERILOG) "\
+	$::env(PDK_ROOT)/$::env(PDK)/libs.ref/sg13g2_io/verilog/sg13g2_io.v
```
dnltz, line 42:
```suggestion
	$::env(PDK_ROOT)/$::env(PDK)/libs.ref/sg13g2_io/verilog/sg13g2_io.v \
```

ihp-sg13g2/libs.ref/sg13g2_stdcell/lef/sg13g2_tech.lef
```diff
@@ -127,7 +127,7 @@ LAYER Metal2
   WIDTH 0.39        0.21    0.24    0.24
   WIDTH 10.0        0.21    0.24    0.60 ;
   MINIMUMCUT 2 WIDTH 1.4 ;
-  AREA      0.144 ;
+  AREA      0.18 ;
```
dnltz: This is an issue with OpenROAD and should be fixed there. I would suggest to add a comment mentioning why you add this hack and also add the original value.

ihp-sg13g2/libs.ref/sg13g2_io/lef/sg13g2_io.lef
```diff
@@ -29,7 +29,7 @@ END sg13g2_ioSite
 
 MACRO sg13g2_Corner
     CLASS PAD SPACER ;
-    ORIGIN 180.000 0.000 ;
+    ORIGIN 0.000 0.000 ;
```
dnltz, line 32: @b10346 I saw you changed this value from `0` to `180` on the 10th of March. Can you clarify why you changed it? Let's not change this value once per quarter ;)

ihp-sg13g2/libs.ref/sg13g2_stdcell/lef/sg13g2_tech.lef
mole99: I added a comment to the PR and the commit message. As soon as this gets fixed upstream I'm going to revert this hack. We should both be notified once the issue in ORFS is closed. I don't think a comment is necessary as the old value is saved in the git history.

ihp-sg13g2/libs.ref/sg13g2_io/lef/sg13g2_io.lef
mole99: Please also note @smunaut 's commit message: https://github.com/IHP-GmbH/IHP-Open-PDK/pull/540/commits/982c67c3dca5c8d4e014561982417c0271c1bb86

mole99:
Thanks for the review @dnltz!

ihp-sg13g2/libs.ref/sg13g2_io/lef/sg13g2_io.lef
mole99: Pinging @b10346 again.
But I think we're all in agreement that an offset of 0,0 is the proper way to go. Both Tiny Tapeout and Greyhound were built using this LEF file.

dnltz: Yes, just curious why this got changed again.

mole99: The main reason why the offset is changed back to 0,0 is because a previous version of magic was confused by the offset. This has been fixed in version 8.3.525. However, there's really no reason for an offset other than 0,0 which is why this is changed back.

You would need to ask @b10346 why this was changed in the first place, but this seems to be the proper way.

mole99:
@KrzysztofHerman @sergeiandreyev Could you take a look at merging this?
I'd love to get this in before FSiC so I can create pre-release in ciel and let people try out LibreLane.

mole99:
@KrzysztofHerman @sergeiandreyev I've updated the LibreLane setup to work with the latest stdcell changes in #553. (Edit: fixed PR)

There was a small typo in the Stdcell update 2 that led to these warning:

```
[12:02:09] WARNING  [ORD-2001] LEF macro sg13g2_sdfrbp_1 pin QN missing from liberty cell.
[12:02:09] WARNING  [ORD-2001] LEF macro sg13g2_sdfrbp_2 pin QN missing from liberty cell.

[12:02:57] WARNING  [ODB-0286] Terminal Q_N of CDL master sg13g2_sdfrbp_2 not found in LEF.
[12:02:57] WARNING  [ODB-0286] Terminal Q_N of CDL master sg13g2_sdfrbp_1 not found in LEF.
```

I fixed it in the last commit to this PR: a9162542aaccccf68a3a6c34cefe663625659358
This PR is now ready to merge.

sergeiandreyev:
> I fixed it in the last commit to this PR: https://github.com/IHP-GmbH/IHP-Open-PDK/commit/a9162542aaccccf68a3a6c34cefe663625659358

Hi @b10346, please take a look as well
@mole99, you meant this one #553, am I right?

mole99:
Yes, my bad!

mole99:
I'd like to add that the GDS and the LIB also show `Q_N` as the correct pin name, this fix should therefore be correct.
It would be great if this PR can be merged soon.

sergeiandreyev:
hm, I see `QN` pin in layout:
image

mole99:
@sergeiandreyev Wait, you're right! It seems I got that mixed up.
So the error is in the GDS, and the LEF is most likely generated from the GDS, so the error is there too.
The Verilog, Spice, CDL and LIB all use `Q_N` for all of the cells.

I fixed the GDS in the latest commit.

mole99:
I removed the last commit in favor of #565.

KrzysztofHerman (approved):

sergeiandreyev:
I'm going to rerun all the failing LVS regression tests, to check if these are actually clean or not

mole99:
Ping @sergeiandreyev, it's all clean :)

sergeiandreyev:
ok, one final thing, before Alexey will write down his feedback - did you check the GDS of IO cells after your `sg13g2_io.lef` update? or better to say, did you try to run some designs with IOs - is the `Corner` cell placed correctly?

mole99:
@sergeiandreyev Yes, both [Greyhound](https://github.com/mole99/greyhound-ihp/tree/main) and the latest two Tiny Tapeout chips on IHP use this setup.

ihp-sg13g2/libs.ref/sg13g2_io/lef/sg13g2_io.lef
b10346: > @b10346 I saw you changed this value from `0` to `180` on the 10th of March. Can you clarify why you changed it? Let's not change this value once per quarter ;)

Hi, yes I can remind ;). 
We generated new LEF on March with following parameters: ORIGIN 0.000 0.000 ; FOREIGN 0.000 0.000.
Initial LEF had following: ORIGIN 180.000 0.000; FOREIGN -180.000 0.000 ;
We also left only one corner in GDS assembly (it was two before, the same cells with different placement orientation).
Corner had problems with correct placement after this change. After changing ORIGIN back to  180.000 0.000. Flow worked correctly. You can find the history here: https://github.com/IHP-GmbH/IHP-Open-PDK/pull/387

My opinion regarding current changes is following:
1) ORIGIN and FOREIGN coordinates have to be zeros always for all library cells.
2) If we apply this change, it is needed to check that GDS layout does not require any changing of cell orientation and final GDS layout of design is correct.

PR #563, opened 2025-06-27
Added bulk node to resistor models rhigh + rppd + rsil cont. 

KrzysztofHerman:
This is a continuation of the original PR #520 developed by @p-fath. 
 - symbols were changed to have a 3-rd terminal exposed as a device property
 - modification in DC and MC examples in `sg13g2_tests/` directory
 - support for Xyce `r3_cmc` added  (need plugin!)

image

Known issues:
     - diferrences between ngspice and xyce simulation results;
     - LVS is crashing since the LVS deck does not recognize resistors as       3 terminal devices.


KrzysztofHerman:
@dwarning, @p-fath this update includes a fix, which solves [this](https://github.com/Xyce/Xyce/issues/150) issue reported on Xyce github. It was contributed  as #568 and added to this PR. 
It requires `R3_CMC` model regeneration using `ADMS`

p-fath:
Really thanks, this works great! I tested the ngspice and xyce implementations. 

- Works fine with ngspice!
- @KrzysztofHerman  In ihp-sg13g2/libs.tech/xyce/models/resistors_mod.lib line 195-197 should be (like it's done in the other resistors):
+nsmm_rsh=1 ; number of standard deviations for rsh
+nsmm_w=1  ; number of standard deviations for w
+nsmm_l=1   ; number of standard deviations for l
- Unfortunately, Xyce is more than 50% off with the new implementation using the r3_cmc models

Issue #547, opened 2025-06-17
Problem With No Filler stack


PhillipRambo:
When i use the NoFillerStack Pcell in klayout i experience that the NoFill perimieter for Active layer and Gate Poly, is not excluded, meaning there is applied fillers in there. For the metal layers it works but not for the ones described above.

![Image](https://github.com/user-attachments/assets/cf8c236f-32cb-4749-bb5e-20b037a8ab2a)
![Image](https://github.com/user-attachments/assets/2e3dacf6-c8ec-4db9-8c7f-eff89f623f12)

KrzysztofHerman:
@PhillipRambo please verify this on your end

PhillipRambo:
@KrzysztofHerman confirmed!

PR #564, opened 2025-06-27
Changes in Act and GatPoly filler macro. No_Fill layers were added to

KrzysztofHerman:
Changes in Act and GatPoly filler macro `sg13g2_filler_ActGatP.lym`. no_fill layers were added to the exclusion areas.
It solves #547 issue

image


sergeiandreyev (approved):

PhillipRambo:
I think this issue should be opened again. I found issue with the active and gatpoly filler in the case where i use Pwell.block in my layout. The DRC clearly states that the min distance to the Pwell block should be min 1.50um. Its not directly related to the exclusion layer, so maybe this should be a seperate issues.:

image

image

PR #575, opened 2025-07-05
A KLayout D25 view setup

klayoutmatthias:
Does not fix an issue, but provides an enhancement. Does not add intrinsic features or code, hence no test.

Adds "basic.lyd25" which provides a 3d setup of the basic stack (Poly to M7, including MIM). Includes diffusion and nwell. As TinyTapeout has a 3d view, the PDK should have one too :)

I don't know where to edit README. I would love to do so, if you give me a hint.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/d25/basic.lyd25
+++ b/ihp-sg13g2/libs.tech/klayout/tech/d25/basic.lyd25
@@ -1,0 +1,118 @@
+<?xml version="1.0" encoding="utf-8"?>
+<klayout-macro>
+ <description/>
+ <version/>
+ <category>d25</category>
+ <prolog/>
+ <epilog/>
+ <doc/>
+ <autorun>false</autorun>
+ <autorun-early>false</autorun-early>
+ <priority>0</priority>
+ <shortcut/>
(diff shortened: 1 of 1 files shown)


mole99:
Very nice! I just want to mention that there is a D25 setup in the `dev` branch: https://github.com/IHP-GmbH/IHP-Open-PDK/blob/dev/ihp-sg13g2/libs.tech/klayout/tech/d25/sg13g2_beol.lyd25

But your approach seems very clean :)
Maybe both can be merged together?

PS: You need to open the PR to `dev` and sign-off your commits using `git commit -s`.

klayoutmatthias:
Ah, I see. You're using Git flow. Sorry, I did not notice. But it makes sense.

The dev version is fine. I just used variables for thickness and deep mode for the boolean computations (feeling safer in terms for full-chip performance). If you want to adopt this, make sure you switch back to flat after the computation as d25 does not like deep mode currently.

And thanks for the notice about signed commits. 

I'll close this PR as it is not substantially different from dev and after all, I was just interested in benchmarking with the TT view :)

Best regards,

Matthias

klayoutmatthias:
Plus dev is substantially different - it has colors :)

Matthias

PR #598, opened 2025-07-16
Introduce Tox

dnltz:
This PR introduces tox as tool to manage common tasks. It only implements lint checks for now but could be extended to also LVS tests in the future.

See https://tox.wiki/en/latest/index.html


KrzysztofHerman:
Dear @dnltz we have converted this PR into a draft since we have to focus first on the features directly related to the PDK.

PR #600, opened 2025-07-16
nmos_cl symbols for xschem updated

### LINKED ISSUE #594: Polarity in ESD-Devices is wrong in LVS
Hi everybody!

Xschem:

Image

Netlist:

```
.SUBCKT TESTESD VDD
*.PININFO VDD:B
D1 sub! VDD nmoscl_4 m=2
.ENDS
.GLOBAL sub!

```
Klayout LVS:

Image

Files: (Xschem+Klayout)

[TESTESD.zip](https://github.com/user-attachments/files/21229534/TESTESD.zip)

Extracted netlist:
```
.SUBCKT TESTESD VDD SUB!
D$1 SUB! VDD nmoscl_4 m=2
.ENDS TESTESD
```

Description:
Although the test for this test case is LVS pass, but it says that the pins should be flipped.
In bigger circuit would the test fail. Here it is OK because we do not have any other components tied to these nets.
Please consider that by netlist generation / LVS extraction to be repaired.

Thank you in advance.
Ghaith.

KrzysztofHerman:
Fixes #594 
Port order changed for `xschem` symbols.

### DIFF
--- a/ihp-sg13g2/libs.tech/xschem/sg13g2_pr/nmoscl_2.sym
+++ b/ihp-sg13g2/libs.tech/xschem/sg13g2_pr/nmoscl_2.sym
@@ -30,8 +30,8 @@ E {}
 L 4 0 -30 0 -5 {}
 L 4 0 5 0 30 {}
 L 4 -10 -5 10 -5 {}
-B 5 -2.5 -32.5 2.5 -27.5 {name=vdd dir=inout pinnumber=2}
-B 5 -2.5 27.5 2.5 32.5 {name=vss dir=inout propag=1}
+B 5 -2.5 27.5 2.5 32.5 {name=vss dir=inout sim_pinnumber=1}
+B 5 -2.5 -32.5 2.5 -27.5 {name=vdd dir=inout sim_pinnumber=2}
 P 4 4 0 -5 -10 5 10 5 0 -5 {fill=true}
 T {@name} 10 -7.5 2 1 0.2 0.2 {}
 T {@model} 10 2.5 2 1 0.2 0.2 {}
--- a/ihp-sg13g2/libs.tech/xschem/sg13g2_pr/nmoscl_2.sym
+++ b/ihp-sg13g2/libs.tech/xschem/sg13g2_pr/nmoscl_2.sym
@@ -30,8 +30,8 @@ E {}
 L 4 0 -30 0 -5 {}
 L 4 0 5 0 30 {}
 L 4 -10 -5 10 -5 {}
-B 5 -2.5 27.5 2.5 32.5 {name=vss dir=inout sim_pinnumber=1}
-B 5 -2.5 -32.5 2.5 -27.5 {name=vdd dir=inout sim_pinnumber=2}
+B 5 -2.5 27.5 2.5 32.5 {name=vss dir=inout sim_pinnumber=2}
+B 5 -2.5 -32.5 2.5 -27.5 {name=vdd dir=inout sim_pinnumber=1}
 P 4 4 0 -5 -10 5 10 5 0 -5 {fill=true}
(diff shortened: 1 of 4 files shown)


sergeiandreyev (approved):

d-m-bailey:
@KrzysztofHerman Looking at the [xschem documentation here](https://xschem.sourceforge.io/stefan/xschem_man/xschem_man.pdf), it appears that the `pinnumber` attribute is only used for packaged devices (`tEDAx` netlists, see page 54). For spice netlists, the attribute is `sim_pinnumber`, but that only works if all pins have a `sim_pinnumber` attribute (as far as I understand). 
The `propag` attribute (explained on page 56), specifies highlighting through the device. Originally, the second pin (index 1) had `propag=1` which would do nothing. With this change, when the net connected to the first pin `vdd` (index 0) is highlighted, the highlighting will be propagated to the pin with index 1, `vss`. I don't think this wanted.

I suggest removing the `pinnumber` and `propag` properties.

Note, I haven't tested any of the above, but that is my understanding from the documetation.

sergeiandreyev:
Hi @KrzysztofHerman, could you please review the feedback from Mitch?

KrzysztofHerman:
@d-m-bailey thank you for this observation. I have removed  `pinnumber` and `propag` properties and introduced `sim_pinnumber` which is reasonable to make explicit pin ordering.

sergeiandreyev (changes requested):
the update does not impact the testcase attached to the original issue #594

d-m-bailey:
@KrzysztofHerman @sergeiandreyev 
The original commit `642056cb456df764742a2bcefe381430f7fea37d` corrected the problem by reordering the terminals in the `*.sym` files. However, the instances did contain incorrect unused parameters `propag` and `pinnumber`.

Unfortunately, the correction of these parameters to `sim_pinnumber` put the terminals back in the original "incorrect" order. I say incorrect because the terminal order does not match the [spice model subckt order](https://github.com/IHP-GmbH/IHP-Open-PDK/blob/dev/ihp-sg13g2/libs.tech/ngspice/models/sg13g2_moslv_mod.lib). 
```
* ESD MOS Clamp devices
*****************************************************************
.subckt nmoscl_2  VDD VSS
+l=0.36u w=168u m=1 ng=1 trise=0  as = 58e-12  ad = 58e-12 pd = 336u ps = 336u rfmode=0 pre_layout=1

.include sg13g2_moslv_parm.lib

 Nsg13_lv_pmos VDD VSS VSS VSS nmoscl_psp w='w' l='l' nf='ng' as='as' ad='ad' pd='pd' ps='ps' mult='m'  dta=trise ngcon=2
.ends nmoscl_2

.subckt nmoscl_4  VDD VSS
+ l=0.36u w=168u m=1 ng=1 trise=0 as = 114e-12  ad = 114e-12 pd = 672u ps = 672u rfmode=0 pre_layout=1
.include sg13g2_moslv_parm.lib

Nsg13_lv_pmos VDD VSS VSS VSS nmoscl_psp w='w' l='l' nf='ng' as='as' ad='ad' pd='pd' ps='ps' mult='m'  dta=trise ngcon=2
.ends nmoscl_4
```

Please note that the ngspice subckt model order is not the expected order (ie. anode cathode) for normal diode devices. I suppose it's too late to "correct" this by swapping the ports in the ngspice model, right?

KrzysztofHerman:
@d-m-bailey you are completely right as for the `sim_pinnumber`. It does not follow the convention `anode(+) cathode(-)` of a diode. Here I dropped the screenshot from the testcase:

image

So as you can see the pins `net1` and `GND` of the `XD1` instance bind to `VDD`  and `VSS` respectively giving you the following output of a reversed biased "diode". 

image

sergeiandreyev (approved):

PR #608, opened 2025-07-21
Various fixes for the magic tech file

smunaut:
This PR includes several fixes for the magic tech file.
Created as a draft until @RTimothyEdwards gets a chance to review / validate those.

### DIFF
--- a/ihp-sg13g2/libs.tech/magic/ihp-sg13g2.tech
+++ b/ihp-sg13g2/libs.tech/magic/ihp-sg13g2.tech
@@ -4450,8 +4450,6 @@ variants *
 
  width *mimcap 1140 "MiM cap width < %d (MIM.a)"
  spacing *mimcap *mimcap 600 touching_ok "MiM cap spacing < %d (MIM.b)"
- surround *mimcc *mimcap 260 absence_illegal \
-	"MiM cap must surround MiM cap contact by %d (MIM.d - TV1.c)"
  rect_only *mimcap "MiM cap must be rectangular (Grid Rules)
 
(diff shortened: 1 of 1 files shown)


smunaut:
@RTimothyEdwards Ah well, I thought the commit message was a pretty good summary :sweat_smile: 

The mimcc will generate cuts using that rules `squares-grid 360 420 420`, so as far as I understand that means there will already be a 0.36u border all around the generated cuts in both the involved metals.

That 0.36u corresponds exactly to the `MIM.d` rule which is the MIM enclosure of the `mimcc`. That means that first surround rule serve no purpose. And you can see that rule had something weird in it because it refers to `MIM.d - TV1.c` but `TV1.c` is Metal5 enclosure of `TopVia1` and that rule plays no role in MIM caps.

The minimum contact width is then 360 (border) + 420 (cut) + 360 (border) which is 1.14u. The rule seem to have been copied from the `TopVia1` rule but not adapted for MIM contacts which have a larger required enclosure of the lower layer.

The spacing rule doesn't matter since the border is so wide, 2*360 > 420 ...

The Metal6 overlap was testing metal5 (typo).
And same for the `exact_overlap` rule :  looks like a typo since there is no `m5` involved in MIM

RTimothyEdwards:
@smunaut :  The commit messages were probably sufficient but I didn't investigate each one individually.  The quick summary above really helps.

RTimothyEdwards:
Should be good to go ahead and take out of draft mode.

smunaut:
@RTimothyEdwards Thanks !

PR #619, opened 2025-07-28
sealring.py: Add offset options

jpf91:
Hi all,

When using the sealring with ORFS digital designs, the design is located at (0,0) and the sealring needs to be moved.

(At least this is what the [masked-aes-tapeout](https://github.com/HEP-Alliance/masked-aes-tapeout) README suggests. I used that repo as a reference, as suggested in the Digital Design Workshop in February. I hope this is still the up-to-date way to do things :smile: )

This commit implements additional offset parameters in the `sealring.py` script. Using these parameters the final sealring can be generated directly to avoid the manual moving step.

BTW: There's also an issue in this line:
```python
pcell = layout.add_pcell_variant(lib, pcell_decl.id(), {'w': f'{width}u', 'l': f'{heigth}u'})
```

It actually confuses width and height. But should this be fixed in the script or in the PyCell? The Pycell talks about **w**idth and **l**ength and explicitly mentions that width is on the Y axis. So it is correct, but the definition is at least confusing for me.

Note: The 2 commits here should be squashed, but IIRC github can do this when merging?

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/scripts/sealring.py
+++ b/ihp-sg13g2/libs.tech/klayout/tech/scripts/sealring.py
@@ -15,7 +15,7 @@ import klayout.db
 LIB = 'SG13_dev'
 PCELL = 'sealring'
 
-def generate_sealring(width: float, heigth: float, output: str):
+def generate_sealring(width: float, heigth: float, output: str, offset_x: float, offset_y: float):
     """Function to create a new layout, add the sealring PCell to sealring_top
     and save it somewhere on the filesystem.
 
@@ -25,6 +25,10 @@ def generate_sealring(width: float, heigth: float, output: str):
     :type heigth: float
     :param output: Path and name of the file where the sealring should be written to.
     :type output: str
+    :param offset_x: Translation in X direction in µm.
+    :type offset_x: float
+    :param offset_y: Translation in Y direction in µm.
+    :type offset_y: float
 
     """
     layout = klayout.db.Layout(True)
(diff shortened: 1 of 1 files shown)


KrzysztofHerman:
@jpf91 thank you for the contribution. 
After reviewing the details I have the following feedback:

1. Since we follow so called `gitflow` of development external contributors are not allowed to push the changes into `main` branch, use `dev` instead.
2. The PyCells we deliver in the open-source PDK are also used by proprietary tools, We have to check the consistency of this solution across all the tools we support.
3. Since in ORFS the seal ring is directly passed to klayout while GDS streaming. Maybe a solution to provide automated flow would be an additional tiny script (or additional functionalities in `def2stream.py`), which will do the positioning. 
```
exec klayout -zz -rd design_name=$::env(DESIGN_NAME) \
           -rd in_def=$::env(RESULTS_DIR)/6_final.def \
           -rd in_files="$::env(GDSOAS_FILES) $::env(WRAPPED_GDSOAS)" \
           -rd config_file=$fill_config \
           -rd seal_file=$seal_gds \
           -rd out_file=$::env(RESULTS_DIR)/6_final.$::env(STREAM_SYSTEM_EXT) \
           -rd tech_file=$::env(OBJECTS_DIR)/klayout.lyt \
           -rm $::env(UTILS_DIR)/def2stream.py
```

jpf91:
Thanks @KrzysztofHerman ! I cleaned up the code a bit and squashed the commits.

1. Rebased onto `dev`
2. I see, that's what I thought. So it's probably easier to fix `sealring.py` instead. Do we have to be backwards compatible there? It might break user designs when they swapped height/width manually as a workaround.  Anyway, if you want I'll open a new bug report. In the end it's only a nuisance, nothing important.
3. I see. That's also a nice / more automated solution, but I don't really know how to adjust the ORFS flow :-) I also do not know if there's a certain offset that we always want to use or whether it should be configurable. I just took the 60um from masked-aes-tapeout for my design and that seems to work.

KrzysztofHerman:
@jpf91 please go to [OpenROAD](https://matrix.to/#/#openroad:fossi-chat.org) channel and ask there about how to integrate this change in ORFS flow. In my opinion is it something related to the particular flow. I will keep this PR open meanwhile.

KrzysztofHerman:
@jpf91 after internal discussion the decision is not to update the pycell itself. The positioning of the cell should be handled by the flow. I believe that OpenROAD team will address this feature request.

jpf91:
@KrzysztofHerman  thanks for checking.

I just wanted to add that technically, this proposed change was not in the pycell but rather a helper script instantiating the pycell.

But I understand you'd prefer having this in ORFS instead. I'll raise an issue there and for now just use a custom fork of the script.

sergeiandreyev:
> @KrzysztofHerman thanks for checking.
> 
> I just wanted to add that technically, this proposed change was not in the pycell but rather a helper script instantiating the pycell.
> 
> But I understand you'd prefer having this in ORFS instead. I'll raise an issue there and for now just use a custom fork of the script.

yes, it was a misunderstanding - we will discuss this internally once more
@KrzysztofHerman, FYI

jpf91:
Thanks for discussion this again and merging!

PR #621, opened 2025-07-31
Fix annotation of npn transistors

p-fath:
Annotation of npn transistors is no longer possible; this issue is fixed in this PR.


KrzysztofHerman (approved):

KrzysztofHerman:
@p-fath thank you for the fix.

PR #630, opened 2025-08-06
xschem symbols for padcell library sg13g2_io added; this enables top-level simulation including padframe.

p-fath:
- [ ] Xschem symbols for sg13g2_io created; based on libs.ref/sg13g2_io/sg13g2_io.spi spice netlist file.
- [ ] Enables realization of padframe in xschem
- [ ] Enables simulation of top-level, including padframe (quite important)
- [ ] Added a testbench to simulate the pads (not tested with Xyce!)
- [ ] Xschem Symbols for tri-state pad cells (bidirectional pad cells) not created, still missing!
- [ ] LVS check of an exemplary padframe created with these pad symbols has not been tested yet.
grafik


KrzysztofHerman:
@p-fath I have reviewed this PR and I have the following observations:

1. We are working on our own hierarchical version of the IO cells, this one is top level binding directly to the spice netlist 
2. Bond pad is not a part of IO cell, we have a separate symbol for a bond pad 
3. I like very much the drawings of the schematics
4. The position of the IO ports have to be consistent with the proprietary PDK to facilitate future conversion of the views. 

i have converted it into a draft. If you like we could work together on the hierarchical version.  I can create a separate branch for this in my own fork of the PDK.

p-fath:
@KrzysztofHerman 
If the pins of the schematics of the hierarchical version of the IO cells will match the pins of the .spice netlist file, the proposed symbols will also work fine. 

We’re quite time-limited on our side, but we’d like to help a bit if it’s not too much overhead. 

A separate branch in your fork for the hierarchical version sounds like a good idea!

simi1505:
I am currently working on a top-level simulation, and I would use the same flow as @p-fath. Any information on whether this PR will be merged soon? Thanks!

PR #644, opened 2025-08-19
Flag port mismatch as LVS error

### LINKED ISSUE #484: LVS sees wrong placed net-label but thinks it dosent matter...
i have a miss placed label:

Image

and LVS sees it, but shows all in the greenest green possible:

Image

the files:

[LVS_WTF1.zip](https://github.com/user-attachments/files/20138855/LVS_WTF1.zip)

i use KLayout o.29.12 and the PDKs DEV branch

H-Ojiro:
Fixes #484
Requires klayout 0.30.2 or later
Remove redundant compare

- [x] Tests pass
- [x] Appropriate changes to README are included in PR (No changes necessary)


H-Ojiro:
@KrzysztofHerman 
@sergeiandreyev 

Please send artifacts from failed LVS Regression Testing

@d-m-bailey

sergeiandreyev:
Hi @H-Ojiro, attaching log files:
[logs_43748907073.zip](https://github.com/user-attachments/files/21851674/logs_43748907073.zip)
@FaragElsayed2, do we have anything else that could be shared to analyze the regression failures?..

d-m-bailey:
@sergeiandreyev I'm working with @H-Ojiro on the updates. The log files attached show us what cells are failing but not why. Will try to reproduce the results with the given inputs and report back.

sergeiandreyev:
a comment from my side.. I checked on a couple of standalone testcases and it is working as expected

d-m-bailey:
@sergeiandreyev Thanks. That explains it. We'll make a list of the port difference.

The change flags ports that don't match. Strictly speaking, it flags schematic ports that don't match the layout ports. Layout ports that are missing in the schematic are not flagged.

FaragElsayed2:
@d-m-bailey It would be great to have this as a run option that users can enable or disable based on their preference.

sergeiandreyev:
at least by default it should be switched-on (but I'm also ok to have it hidden from the user, I don't know of any testcase where it should be not enabled to see these violations..)

d-m-bailey:
@FaragElsayed2 That's a good suggestion. I'll talk with @H-Ojiro about implementing a flag to ignore ports.

d-m-bailey:
I looked into the error on the `buf_1` cell. It appears that it's a text layer problem.

`VDD` and `VSS` are on layout metal1/text (8/25), but `A` and `X` are on metal1/label (8/1). metal1/label is not used by the current klayout rules. The `libs.ref/sg13g2_stdcells/gds` layout has all the text on 8/25, so it appears that it's just the pdk test data that may be out of date.

スクリーンショット 2025-08-20 0 10 18


We can update the test data along with the program changes.

sergeiandreyev:
> We can update the test data along with the program changes.

yes, we should not have any `.label` layers in stdcells - a while ago we changed to use `.text` layers for the pin labels

d-m-bailey:
@sergeiandreyev I looked at the test data that the CI runs on and it appears that the test data is older than the actual standard cell data.

@FaragElsayed2 for each standard cell in the test data, it looks like there are 3 versions - normal, digital isolation, and deep nwell isolation. Are these cells generated by a script? 
Would it be possible to generate this test data directly from the most recent standard cell gds data every time the CI runs?

H-Ojiro:
Added --allow_missing_ports option

sergeiandreyev:
> Would it be possible to generate this test data directly from the most recent standard cell gds data every time the CI runs?

It would be very nice to have indeed 👍

hpretl:
That is a super important addition (I would call it a fix) for the LVS :-) I hope you can merge soon.

d-m-bailey:
@hpretl I'm reviewing the automatic test case generation script now. Should be completed this weekend.

d-m-bailey:
@H-Ojiro and I have added a program that recreates the test cells from the pdk libraries. This works for test cells in the libraries, but other cells (like `sg13g2_inv2_and2_flat` and `sg13g2_inv2_and2_deep`) may have to be modified manually.

This change enables the klayout 0.30.2 to fail LVS if any of the source ports are not in the layout.

There is an option ~~`--allow_missing_ports`~~ -> `--allow_unmatched_ports` that retains the pre-change behavior of allowing port name mismatches.

sergeiandreyev:
@d-m-bailey, great! so we're ready to merge?

d-m-bailey:
@sergeiandreyev We're ready.

Do you want the maintainers at mabrains to approve it? @FaragElsayed2

sergeiandreyev:
I think @FaragElsayed2 has likely reviewed this PR, though a second check on the final updates would definitely be helpful..

d-m-bailey:
@FaragElsayed2 comments?

FaragElsayed2:
@d-m-bailey
Looks perfect to me, thanks for the update! It would be great if this switch could also be supported in the GUI run.

Also, please update the README.md to include information about this switch in the [Usage section](https://github.com/IHP-GmbH/IHP-Open-PDK/tree/dev/ihp-sg13g2/libs.tech/klayout/tech/lvs#usage)

@atorkmabrains
Any thoughts?

d-m-bailey:
@FaragElsayed2 

> It would be great if this switch could also be supported in the GUI run.

Excellent point. @H-Ojiro Could you look into this?

H-Ojiro:
@FaragElsayed2   CC: @d-m-bailey 

Added GUI option "Allow Unmatched Ports" at the last PR
Could you take a look at it?

image
 

Please also note that the checking Klayout version is updated to 0.30.2 required by 'flat_missing_ports' command

sergeiandreyev (approved):

PR #673, opened 2025-09-02
klayout(metal fill): fix minimum distance check for metal layers

tskaar:
Currently, the input checks for metal filler distances are all against `distances['distance_m1']`.

This commit fixes the checks to validate each respective metal layer's distance.


sergeiandreyev:
Hi @dnltz, could you please take a look?

dnltz (approved):
Thanks, that is correct.

PR #681, opened 2025-09-11
Change spice netlist prefix for extraced circuits

### LINKED ISSUE #680: Simulatable SPICE netlist from KLayout LVS extraction
## Error
When running LVS-checks with KLayout the extracted is invalid and cannot further be used by a spice simulator, e.g. ngspice.
Trying to use it results in the following error:
```
could not find a valid modelname
    Simulation interrupted due to error!
```

## Reproducability
1. Create an extraced spice netlist by running the LVS testcases
```
make test-LVS-cells
```
2. Create a minimal spice example, which uses the extracted circuit:
```
.lib cornerMOSlv.lib mos_tt
.include sg13g2_inv_1_extracted.cir
Xinv VDD VSS A Y sg13g2_inv_1
```

## Details
The extracted circuit describes the transistors with the `M` prefix, even though they are defined as subcircuits by `libs.tech/ngspice/models/sg13g2_moslv_mod.lib`, so the `X` prefix should be used.

FlinkbaumFAU:
Devices such as sg13_lv_nmos are defined as `.subckt`, not as models. This means the prefix for the spice netlist should be X, so that the extracted netlist can be used for simulation.

Fixes #680 

- [x] Tests pass


sergeiandreyev:
Hi @FaragElsayed2, I guess this could not be approved because of the way the LVS in KLayout is implemented (CDL-based), please confirm..

d-m-bailey:
@sergeiandreyev I think this was the focus of our discussion with @RTimothyEdwards last month.

I thought the conclusion was that xschem and other schematic entry tools would generate 2 netlists, an ngspice compatible simulation netlist (with mosfets etc. as `X` devices) and a CDL netlist for LVS (with mosfets as `M` devices). 

This will probably require most xschem device models to have 2 formats, (`format` for simulation and `lvs_format` for LVS).

There is a `spiceprefix` variable, but we might need an `lvs_spiceprefix` too.

atorkmabrains:
@sergeiandreyev and @KrzysztofHerman Could you please force LVS Regression to rerun as I need to review the logs?

One note, such change will change from the normal flow of using CDL/LVS based netlist to simulation based netlist, Is that what you want?

sergeiandreyev:
> One note, such change will change from the normal flow of using CDL/LVS based netlist to simulation based netlist, Is that what you want?

well, no - that's why I commented that in my view we should reject this (message from September):
> Hi @FaragElsayed2, I guess this could not be approved because of the way the LVS in KLayout is implemented (CDL-based), please confirm..

I just wanted to confirm with you

atorkmabrains:
> > One note, such change will change from the normal flow of using CDL/LVS based netlist to simulation based netlist, Is that what you want?
> 
> 
> 
> well, no - that's why I commented that in my view we should reject this (message from September):
> 
> > Hi @FaragElsayed2, I guess this could not be approved because of the way the LVS in KLayout is implemented (CDL-based), please confirm..
> 
> 
> 
> I just wanted to confirm with you

If the decision to comply with CDL/LVS kind of netlist, then yes, that should be rejected.

PR #671, opened 2025-09-02
KLayout PyCells: Segmented resistor array features

martinjankoehler:
**NOTE:** ~depends on #670, please merge / review this one first, then rebase~

## Features
- existing resistor cells have additional parameters:
   - number of segments
   - segmented connection: serial / parallel / no 

---

- [ ] Tests pass
- [ ] Appropriate changes to README are included in PR

PR #703, opened 2025-10-02
KLayout: Via P(y)Cells

martinjankoehler:
**NOTE:** depends on PRs #671 and #670, please review/merge those first

- PyCell ViaStack supports `Cont`/`Activ`/`GatPoly`
- Native KLayout PCell added separately, which is necessary for the new KLayout `0.30.4` Path-Via-Feature (place via using shortcut 'O')


mole99:
This is a great enhancement to the KLayout setup. I just saw Matthias demonstrate it in the "MICROELECTRONICS 2025: From Technology to Application" webinar.

martinjankoehler:
OK, added some changes:
  - Now enclosure rules `*.c` vs `*.c1` are honored
  - Bigger array spacing rules `*.b` vs `*.b1` are honored

KrzysztofHerman:
@martinjankoehler thank you for these contributions, I am reviewing it. 
The issue I have after merging 670,671 and 703 is that I get these errors after hitting  'O' key:
```
ERROR: /home/herman/github/KrzysztofHerman/IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/python/sg13g2_native_pcell_lib/sg13_klayout_tech_info.py:76: AttributeError: 'ViaInfo' object has no attribute 'klayout_via_type' in PCellDeclaration.via_types
  /home/herman/github/KrzysztofHerman/IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/python/sg13g2_native_pcell_lib/sg13_klayout_tech_info.py:76
  /usr/lib/python3.12/functools.py:995
  /home/herman/github/KrzysztofHerman/IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/python/sg13g2_native_pcell_lib/via_pcell.py:67

```

martinjankoehler:
@KrzysztofHerman 

Thanks for the report, I've added 2 commits that should fix this issue.

KrzysztofHerman:
@martinjankoehler thank you it works, I will continue the review. It looks very good. 
Until now I have one question: What was the motivation to introduce a separate `native` library ?

martinjankoehler:
@KrzysztofHerman this is required for KLayout's `O`-keystroke Feature to work.
I have therefore separated:
1. in regular IHP library 
    - `via_stack`: PyCell based, no KLayout dependency to be compatible with proprietary flow
    - `TechInfo`: encapsulating DRC rules reusable for both `via_stack` and `via_pcell` below
3. KLayout-*native* library, currently containing `via_pcell`
    -  after this library is registered/loaded at startup, then KLayout knows what to do if user presses `O`

KrzysztofHerman:
@martinjankoehler a time ago we have introduced a "preprocessor like" constructs to 
distinguish between different tools, which actually make function calls ie. 
```
#ifdef KLAYOUT
        klayout specific code
#else

#endif
```
maybe it is a helpful information.

martinjankoehler:
> @martinjankoehler a time ago we have introduced a "preprocessor like" constructs to distinguish between different tools, which actually make function calls ie.
> 
> ```
> #ifdef KLAYOUT
>         klayout specific code
> #else
> 
> #endif
> ```
> 
> maybe it is a helpful information.

Thanks for the info, I already saw it. In this case I found it cleaner to keep it separate for multiple reasons:
- cleaner code with separation of concerns
- easier to debug in isolation without preprocessing
- the `native` library can serve as a template for other Open PDKs to implement the `O`-Feature

IMO in the IHP context, these native pcell library should be kept to a minimum (to honor the PyCell4KLayout strategy), but is useful exactly in those rare cases of KLayout-specific features (like `O`), where a real *native* KLayout PCell is mandatory.

hpretl:
@KrzysztofHerman What is the plan for the merge of this PR? We are using this in our branch, so it would be great to get it merged here as well.

martinjankoehler:
@KrzysztofHerman @sergeiandreyev like we discussed today, I've now rebased this branch on `dev` (and removed the guard ring code)

KrzysztofHerman:
@martinjankoehler thank you for this contribution. I have a question. When one switch between the metals the path tool should use the minimal width corresponding to the actually used metal layer. Would it be possible to implement in the current version of klayout or it does need some modifications/extensions in the tool ?

KrzysztofHerman:
@martinjankoehler 
I found another one: In our documentation we have a setup for cadence like [key-bindings](https://ihp-open-pdk-docs.readthedocs.io/en/latest/analog/klayout.html). Shall we remove the line `edit_menu.cell_menu.adjust_cell_origin:O;` to make 
it work ? If not klayout will pop-up a window that the key binding is ambiguous.

martinjankoehler:
> @martinjankoehler I found another one: In our documentation we have a setup for cadence like [key-bindings](https://ihp-open-pdk-docs.readthedocs.io/en/latest/analog/klayout.html). Shall we remove the line `edit_menu.cell_menu.adjust_cell_origin:O;` to make it work ? If not klayout will pop-up a window that the key binding is ambiguous.

Yes, I would suggest removing the ambiguous other binding.

martinjankoehler:
> @martinjankoehler I found another one: In our documentation we have a setup for cadence like [key-bindings](https://ihp-open-pdk-docs.readthedocs.io/en/latest/analog/klayout.html). Shall we remove the line `edit_menu.cell_menu.adjust_cell_origin:O;` to make it work ? If not klayout will pop-up a window that the key binding is ambiguous.

@KrzysztofHerman I've asked @klayoutmatthias on this topic. With the current KLayout infrastructure, the O-Tool has no access to the minimal-width DRC information. 
Therefore always the least recently used width value is used for each layer.

KrzysztofHerman:
@martinjankoehler thank you for this contribution.

PR #720, opened 2025-10-14
Adding DRC regression for G2 cells

FaragElsayed2:
- DRC regression testing for all SG13G2 cell libraries (pr, stdcell, io, and sram)
-  Adding actions for running DRC for all cells

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -174,7 +174,8 @@ cython_debug/
 *.lyrdb
 *.lvsdb
 unit_tests_*
-drc_run_
+drc_run_*
+drc_cells*_
 lvs_run_*
 *_extracted.cir
 cells_tests_*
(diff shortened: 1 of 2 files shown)


sergeiandreyev:
Hi @FaragElsayed2, am I right that we need to review that spreadsheet with DRC errors/waivers before taking action here? There are some cells in DRC regression that fail, should we expect that?

FaragElsayed2:
> Hi @FaragElsayed2, am I right that we need to review that spreadsheet with DRC errors/waivers before taking action here? There are some cells in DRC regression that fail, should we expect that?

Hi @sergeiandreyev 

Yes, those are the failed cells listed in this [sheet](https://docs.google.com/spreadsheets/d/15hQ6v4G-VDdzUnhnS6TAeqFj1ugmf88LFq0cJPtRhIg/edit?gid=0#gid=0). I believe these are actual violations unless the DRM is updated to explicitly waive them for those specific cells.

PR #725, opened 2025-10-22
Gnucap support using Verilog-AMS

pepijndevos:
This PR enables the use of the IHP PDK with the Gnucap simulator, by utilizing [SpiceArmyKnife.jl](https://github.com/NyanCAD/SpiceArmyKnife.jl) to convert the PDK to Verilog-AMS.

Gnucap is a versatile simulator without direct spice lineage/baggage that has a plugin system that is in the process of growing comprehensive Verilog-AMS support far beyond the CMC subset supported by ADMS or even OpenVAF.

SpiceArmyKnife.jl rose from the ashes of CedarSim to use its sophisticated parsers to parse and convert netlist languages. It can currently apply some compatibility transforms between spice dialects and convert a substantial subset of spice to Verilog-A.

(If there is interest, we could investigate if Xyce/Ngspice/Gnucap/... could be built from a unified codebase on the fly to ease maintenance)

This is part of my work on the [NyanCAD](https://nyancad.com/) schematic editor and simulation environment. (Both NyanCAD and Gnucap receive funding from NLnet :heart:)

This PR is not yet complete but has reached a stage where it can replicate the [ngspice example in the docs](https://ihp-open-pdk-docs.readthedocs.io/en/latest/analog/ngspice.html) by running

```
make
gnucap test.gc
```

```
#           ids(XM1.M1)
 27.        1.1456u   
```

The verilog files were generated using
```
julia --project bin/convert.jl /.../IHP-Open-PDK/ihp-sg13g2/libs.tech/xyce/models/cornerMOSlv.lib /.../IHP-Open-PDK/ihp-sg13g2/libs.tech/gnucap/models/cornerMOSlv.va --inpu
t-simulator xyce --output-simulator gnucap
```

:warning: Xyce bug? During conversion I found that while ngspice uses the pspnqs103va model, Xyce uses the psp103_va model, but passes it nqs-only parameters! I have changed the model for a successful conversion, but have not confirmed Xyce does in fact support this model.

Todos from my end before this is ready

- [x] Add readme/docs for how to use this
- [x] Remove hardcoded path
- [x] Convert the rest of the PDK
- [x] Resolve Xyce model situation (change model? remove unsupported parameters?)
- [x] Report mosvar.va compilation failure
- [ ] Run more extensive tests (is there a body of test netlists for regression testing?)


dwarning:
Xyce is using ADMS. And ADMS can't compile the specific psp NQS code. So for IHP PDK the non NQS version is used for Xyce. IMHO minimal deviations are only in highest frequency range.
What was the problem with mosvar compilation?

felix-salfelder:
> What was the problem with mosvar compilation?

The mosvar code has triggered a bug with nested blocks. It is fixed in 859c1a4. Thanks for testing!

pepijndevos:
Another Xyce discrepency: `sg13g2_lv_nmos_vfbo` used in the statistical model but the module itself uses `sg13g2_lv_nmos_vfbo_mm` causing an error when compiled to Verilog which isn't as loosey-goosey about undefined or unsupported parameters.

pepijndevos:
This now generates reasonable looking verilog for the entire pdk that successfully compiles to gnucap plugins, so the next step is documentation and testing.

I could commit the generated verilog or build it on the fly, up to you. Maybe it's nicer for users if the verilog is ready to go and they only need gnucap and gnucap-modelgen-verilog and not spicearmyknife but by that logic you could also commit the C++ code and then you wouldn't even need modelgen.

KrzysztofHerman:
@pepijndevos @felix-salfelder 
I was able to install gnucap and regenerate the models. The basic example works. In order to 
evaluate the quality of results I would like reuse our examples provides in `xschem` directory.
In fact we have a bunch of circuits where you can launch either ngspice or Xyce from the same schematic. 

I am thinking to do something similar to evaluate gnucap but:
- how to do embedded sweeps ?
- how to format the output (CSV) ? 

Could you also prepare a mixed-signal testcase  based on our technology to showcase the potential of gnucap in this domain ?

felix-salfelder:
On Wed, Nov 19, 2025 at 06:09:58AM -0800, Krzysztof Herman wrote:
> KrzysztofHerman left a comment (IHP-GmbH/IHP-Open-PDK#725)
> 
> @pepijndevos @felix-salfelder 
> I was able to install gnucap and regenerate the models. The basic example works. In order to 
> evaluate the quality of results I would like reuse our examples provides in `xschem` directory.
> In fact we have a bunch of circuits where you can launch either ngspice or Xyce from the same schematic. 

Thanks Krzysztof for looking into it.

We have code that supports gEDA/lepton file format, which is good enough
for simulation [1]. Perhaps xschem isn't far off, but I have not studied
it in detail. (We have funding for this work, but I have no time to do
it).

> I am thinking to do something similar to evaluate gnucap but:
> - how to do embedded sweeps ?
> - how to format the output (CSV) ? 

The "dc" command implements nested sweeps. It supports spice style
sources, and also parameters. We allow Spice style sources in an
otherwise Verilog circuit, but the standard does not require it. In fact
the LRM says little about how a simulation should be performed. Feedback
welcome.

The output table format is simplistic for human readibility. Some tools
like Gnuplot understand it as-is. You could add commas to obtain CSV.

I have updated the manual [2] to cover these topics in more detail.

> Could you also prepare a mixed-signal testcase  based on our
> technology to showcase the potential of gnucap in this domain ?

I am not familiar with the PDK. Do you have any mixed-signal models? We
could use them for testing and/or prioritise features you'd need.

NB: This branch is great for testing (thanks Pepijn!), but It needs more
thought. Verilog-AMS has "paramset" which could be used effectively as a
substitute for .model instances. Parameters also have a type and a range.
These could be added for documentation...

[1] http://gnucap.org/dokuwiki/doku.php/gnucap:user:netlist_import_and_export:geda
[2] http://gnucap.org/dokuwiki/doku.php/gnucap:manual:commands:dc

KrzysztofHerman:
> On Wed, Nov 19, 2025 at 06:09:58AM -0800, Krzysztof Herman wrote: KrzysztofHerman left a comment ([IHP-GmbH/IHP-Open-PDK#725](https://github.com/IHP-GmbH/IHP-Open-PDK/pull/725)) @pepijndevos @felix-salfelder I was able to install gnucap and regenerate the models. The basic example works. In order to evaluate the quality of results I would like reuse our examples provides in `xschem` directory. In fact we have a bunch of circuits where you can launch either ngspice or Xyce from the same schematic.
> Thanks Krzysztof for looking into it. We have code that supports gEDA/lepton file format, which is good enough for simulation [1]. Perhaps xschem isn't far off, but I have not studied it in detail. (We have funding for this work, but I have no time to do it).
Good to hear that it will come some day.
> I am thinking to do something similar to evaluate gnucap but: - how to do embedded sweeps ? - how to format the output (CSV) ?
> The "dc" command implements nested sweeps. It supports spice style sources, and also parameters. We allow Spice style sources in an otherwise Verilog circuit, but the standard does not require it. In fact the LRM says little about how a simulation should be performed. Feedback welcome. The output table format is simplistic for human readibility. Some tools like Gnuplot understand it as-is. You could add commas to obtain CSV. I have updated the manual [2] to cover these topics in more detail.

An hour ago I was able to make plots using gnuplot, seems that python will do the job. 

> Could you also prepare a mixed-signal testcase based on our technology to showcase the potential of gnucap in this domain ?
> I am not familiar with the PDK. Do you have any mixed-signal models? We could use them for testing and/or prioritise features you'd need. NB: This branch is great for testing (thanks Pepijn!), but It needs more thought. Verilog-AMS has "paramset" which could be used effectively as a substitute for .model instances. Parameters also have a type and a range. 

Ok, let's see what can we do together. I was thinking about a simple set of transmission gates controlled by a clocked decoder. 

These could be added for documentation... [1] http://gnucap.org/dokuwiki/doku.php/gnucap:user:netlist_import_and_export:geda [2] http://gnucap.org/dokuwiki/doku.php/gnucap:manual:commands:dc

felix-salfelder:
On Thu, Nov 20, 2025 at 02:46:58AM -0800, Krzysztof Herman wrote:
> Ok, let's see what can we do together. I was thinking about a simple
> set of transmission gates controlled by a clocked decoder.

Thanks Krzysztof or looking into this.

Any circuit will be useful for regression testing and documentation. I
am happy to help with examples provided.

With the commits here I could already identify and fix some bugs in
Gnucap. Thanks Pepijn for your patience. Yesterday, I made snapshots
that seem to work as expected and/or could be used as a reference. More
bugs are conceivable in what we have and in what's going to come.

The device models presented could be reformulated to ordinary paramsets,
as this is the mechanism described in the standard for this purpose. A
paramset is a hybrid between a module and a "spice model" instance.
Gnucap and Gnucap-Modelgen-Verilog support it, presumably other
Verilog-AMS implementations also do.

Combining paramsets and behavioural code prior to compilation yields
smaller code and faster simulation, regardless of the simulator. This is
also possible for "module" in principle but not implemented anywhere, to
my knowledge.  (and there a gaps in the standard related to binning).

Paramsets could be also used to run IHP examples with vanilla "cmc"
devices shipped with the "gnucap-models" repo. These will be compiled
regularly for regression testing anyway. With more users involved,
further automation and some collaboration is clearly desirable.

KrzysztofHerman:
@felix-salfelder it iwll take a little bit longer than I taught but thanks you your support and fixes I have reported on Fossi chat I think we can get some examples soon.

KrzysztofHerman:
@felix-salfelder 
Here goes an example of a analog-mixed signal simulation. It is a VerilogAMS `nand` gate followed by a transistor level inverter.

```
// Verilog
// Libraries
verilog
load mgsim
load ./gates.so
load vams/vpulse.so
load $PDK_ROOT/$PDK/libs.tech/gnucap/plugins/psp103_nqs.so
load $PDK_ROOT/$PDK/libs.tech/gnucap/plugins/cornerMOSlv_tt.so

ground gnd;
anand xnand(nand i1 i2);
sg13_lv_nmos #(.w(1.0e-6), .l(0.13e-6), .ng(1)) XM1(xout, nand, vss, vss);
sg13_lv_pmos #(.w(1.0e-6), .l(0.13e-6), .ng(1)) XM2(xout, nand, vdd, vdd);

spice
Vdd1 vdd gnd 1.2
Vss1 vss gnd 0.0
V1 i1 gnd pulse(0, 1.2, 0, 1n, 1n, 5u, 10u)
V2 i2 gnd pulse(0, 1.2, 0, 1n, 1n, 3u, 10u)

.print tran   v(i1) v(i2) v(nand) v(xout)
.tran 0 10u 10n > arun.txt trace off quiet

```

The VerilogAMS code is the following:

```

`include "disciplines.vams"

module anand (out, in1, in2);

input in1, in2; 
output out; 
electrical out;
electrical in1, in2;
parameter real vh = 1.2;			// output electrical in high state
parameter real vl = 0;			// output electrical in low state
parameter real vth = (vh + vl)/2;	// threshold electrical at inputs
parameter real td = 0 from [0:inf);	// delay to start of output transition
parameter real tt = 0 from [0:inf);	// transition time of output signals

analog begin
    @(cross(V(in1) - vth) or cross(V(in2) - vth));

    V(out) <+ transition( !((V(in1) > vth) && (V(in2) > vth)) ? vh : vl, td, tt );
end
endmodule

```

After simulation one can get the plots.
image

I have the following observations:

1. The simulation is slow ~30s
2. What could be the reason of the osculations at the output ?
3. The initial voltage at the `xout` output is 2.6 V. Any explanation for that ?

felix-salfelder:
On Tue, Dec 23, 2025 at 02:58:44PM -0800, Krzysztof Herman wrote:
> @felix-salfelder 
> Here goes an example of a analog-mixed signal simulation. It is a
> VerilogAMS `nand` gate followed by a transistor level inverter.

Thanks Krzysztof.

> 1. The simulation is slow ~30s

The main reason is the time step argument. With the time step omitted,
Gnucap will do it automatically.

.tran 10u > arun.txt trace off quiet

Automatic time stepping can slow down the simulation of small circuits.
Switching it off requires a modified transient command, not a priority
for us right now.

With that, most time is spent in device evaluation, inefficient
derivatives and obnoxiously large models dont play well. (Working on
it.)

Regardless, you should consider

- paramset. compile a smaller model (explained above)
- disable debugging, compile plugins with -DNDEBUG
- optimise, set -O2

Use the "status" command to get an overview of how much time is spent in
which part(s) of the algorithm(s).

Other tweaks may apply to other circuits, let's cross these bridges when
we get there.

> 2. What could be the reason of the osculations at the output ?

Trap ringing is caused by sharp transitions hitting trapezoidal
integration. One way around that is to use another integration method.

Change the default with

`options method=gear

or select the method used in relevant devices. Method selection is
missing in the standard and hence this is Gnucap specific.

sg13_lv_nmos #(.$method("gear") .w(1.0e-6),...) XM1(xout, nand, vss, vss);
sg13_lv_pmos #(.$method("gear") .w(1.0e-6),...) XM2(xout, nand, vdd, vdd);

See [1] for the rules.

Another possibly more portable approach is to use soft(er) transitions.
Some low passing (dont forget to choose gear) or hardcoded smoothing may
do the trick.

> 3. The initial voltage at the `xout` output is 2.6 V. Any explanation for that ?

Probably a numerical issue with incremental updates. As a workaround set

`options vmin = 0
`options vmax = 1.2

These default to -5 and 5 respectively, which is a little outdated. Both
set to 0 should be most failsafe. And I am inclined to recommend this
setting at the cost of a few extra iterations during the initial op analysis.

[1] http://gnucap.org/dokuwiki/doku.php/gnucap:manual:tech:verilog?#hierarchical_system_parameters

felix-salfelder:
There is a bug in the HBT envelope module. May lead to taking the fractional power of a negative number.
[temp.patch](https://github.com/user-attachments/files/24585227/temp.patch) fixes it.

It remains to check where the wrong expressions originate from. $temperature is already Kelvin, and the offset dance is not even needed.

pepijndevos:
the original converts spice `temper` in celcius to kelvin

https://github.com/IHP-GmbH/IHP-Open-PDK/blob/a2bf8ea81aee7d0fcdd6d62168edca0d7d0bcb08/ihp-sg13g2/libs.tech/ngspice/models/sg13g2_hbt_mod.lib#L111

and then my converter converts `temper` to `$temperature` with an offset

https://github.com/NyanCAD/Cadnip.jl/blob/690579232f1e071d9a14f97445a2567a921551a0/SpiceArmyKnife.jl/src/cg_veriloga.jl#L268-L269

so you get a double conversion except they use slightly different offsets, with mine being the exact one.

Changing the offset is easy, avoiding the double conversion would require somehow parsing the whole expression to understand if a double conversion is happening.

KrzysztofHerman:
@pepijndevos I am still evaluating this PR. Could you please take a look at this two examples provided by @felix-salfelder, which take advantage of `paramsets` ?
[sg13_lv_nmos_tt.vams.txt](https://github.com/user-attachments/files/24740027/sg13_lv_nmos_tt.vams.txt)
[sg13_lv_pmos_tt.vams.txt](https://github.com/user-attachments/files/24740028/sg13_lv_pmos_tt.vams.txt)
The goal would be to maintain the same logic we have in the spice library but using as much as possible the Verilog-AMS features.

pepijndevos:
Translating the spice logic to paramset is the first thing I tried but runs into issues because once you translate subcircuits to modules, you can't put a paramset inside a module as the spice code puts a model card inside a subcircuit, so you're forced to "inline" the model card into its instantiation, resulting in the code currently in this PR.

In my opinion this is a fine result since paramsets are basically lightweight modules that support binning (which we're not using)

What Felix seems to have done is flatten this subcircuit plus model card into one paramset. In this reduced case that may be possible with a nontrivial detection and transformation, but not every subcircuit can be transformed in this way because paramsets are rather limited in what they can contain.

So in my opinion trying to combine a subcircuit and modelcard into one paramset is an optimization that may not be worth much in reality.

felix-salfelder:
Sorry Pepijn if I did not explain it sufficiently. Here goes yet another attempt.

"paramsets" map device instances to models according to the standard. This is established practice, and not for discussion. If ever "translating spice logic" has motivated this approach, we should take the lesson, rather than try to reinvent the wheel. One key aspect is the removal of the need for features in model compilers and simulators. It has nothing to do with Gnucap or Modelgen-Verilog specifically. Nor does it have to do with binning.

Most "model compilers" do not understand hierarchy, nor will they optimise it. There is no benefit in and no technical need to compile the hierarchical modules you supplied. paramsets may become clearer to you, once Gnucap will also support osdi blobs that we create from paramsets. (Not from modules, and, unless you fix it, it will stay like this.) NB: osdi already works with ngspice.

Regarding this patch: your modules are not in conflict with the paramsets. Keep the modules if you like. For simplicity you could have them point at the paramsets, not at the generic PSP103VA and clean up the overhead parameters that are already in the paramsets. I can see real use cases for "modules" (once tidied up) but possibly beyond the scope right now.

KrzysztofHerman:
Dear @pepijndevos I will close this PR since we are now developing a full set of libraries which should give support to gnucap and be more aligned with LRM and advanced features of Verilog-AMS language. Nevertheless your pr was an initial spark for starting to support gnucap simulator, thank you for  that.

PR #728, opened 2025-10-23
Add standalone DRC script and update `run_drc.py` to use it

### LINKED ISSUE #688: Top-level DRC entry points
I noticed that the organization of the DRC rule decks has been revised.
Now all DRC rules have their own file, which is great for keeping an overview and tracking changes in git! 👏

However, there is no longer a top-level entry point to the DRC rule sets. This is an unnecessary restriction and should be considered a regression compared to the previous state of `dev`.

KLayout has the ability to include files as if they were pasted there, combining this with some `if` statements for the switches (e.g. `density`, `no_feol`, `no_offgrid` etc.), this can be used to match the functionality of the Python script.

```
# %include to_include.drc
```

See here for more documentation under "Including other files": https://www.klayout.de/doc-qt5/manual/drc_runsets.html

This allows users to run the rulesets right away without having to preprocess the rule deck with a Python script.
In fact, I already talked about this in this issue: https://github.com/IHP-GmbH/IHP-Open-PDK/issues/221

Please recreate `sg13g2_maximal.lydrc` and `sg13g2_minimal.lydrc` (or similarly named runsets) that can be used without preprocessing.

Once this is implemented, I can update LibreLane to the latest version of the `dev` branch.

mole99:
This PR implements a standalone DRC script and updates `run_drc.py` to use it.

This change means that it is no longer necessary to generate temporary DRC decks, and experienced users (or tools) can use the standalone DRC deck directly. This can be useful for debugging purposes, for example.

- Organize the rule decks into subfolders (`feol`, `beol`, `forbidden`, `pin`).
- Add switches to each rule deck.
- Create the top-level DRC deck (`ihp-sg13g2.drc`).
- Update `run_drc.py` to use the top-level DRC deck
  - Remove the `generate_drc_run_template` function, as it is not used anymore.
  - Remove the `--macro_gen` flag, as it is not necessary anymore.

As before, `run_drc.py` works in mp-mode with and without a table list:

```
python3 run_drc.py --path=testing/testcases/unit/activ.gds --run_mode=deep --run_dir=test_activ --no_density --mp 1
python3 run_drc.py --path=testing/testcases/unit/activ.gds --run_mode=deep --run_dir=test_activ --no_density --mp 16
python3 run_drc.py --path=testing/testcases/unit/activ.gds --run_mode=deep --run_dir=test_activ --no_density --table activ --mp 1
python3 run_drc.py --path=testing/testcases/unit/activ.gds --run_mode=deep --run_dir=test_activ --no_density --table activ --mp 16
```

As for whether this may be confusing for users, I don't think so. The usage of `run_drc.py` is well documented: https://github.com/IHP-GmbH/IHP-Open-PDK/blob/dev/ihp-sg13g2/libs.tech/klayout/tech/drc/README.md

This PR allows me to update LibreLane and CACE to the latest PDK version, even before I update both to use the `run_drc.py`.

Fixes #688

@sergeiandreyev  @KrzysztofHerman

@atorkmabrains If you have any questions about the implementation, feel free to ask.

PS: I noticed that `run_drc.py` has a switch for offgrid rules, however that switch is not currently in use.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/drc/ihp-sg13g2.drc
+++ b/ihp-sg13g2/libs.tech/klayout/tech/drc/ihp-sg13g2.drc
@@ -168,6 +168,7 @@ logger.info("CONNECTIVITY_RULES enabled: #{CONNECTIVITY_RULES}.")
 def get_drc_values(logger)
   script_dir = File.expand_path(File.dirname(__FILE__))
   $drc_json_default = $drc_json_default || File.join(script_dir, 'rule_decks/sg13g2_tech_default.json')
+  $drc_json = $drc_json || File.join(script_dir, '../../python/sg13g2_pycell_lib/sg13g2_tech_mod.json')
 
   tech_rules = {}
   if $drc_json && $drc_json != $drc_json_default


ihp-sg13g2/libs.tech/klayout/tech/drc/run_drc.py
```diff
@@ -764,8 +673,14 @@ def run_check_by_flag(flag_enabled: bool, name: str):
...
-    drc_file = generate_drc_run_template(rule_deck_full_path, run_dir, args.table)
-    result_dbs.append(run_check(drc_file, table_name, layout_path, run_dir, switches))
+    if table_name != "main":
+        # Disable all runset switches
+        # since only a single table is enabled
+        switches["no_feol"] = "true"
+        switches["no_beol"] = "true"
+        switches["no_forbidden"] = "true"
+        switches["no_pin"] = "true"
+    result_dbs.append(run_check("ihp-sg13g2.drc", table_name, layout_path, run_dir, switches))
```
dnltz: This line assumes the script runs in `ihp-sg13g2/libs.tech/klayout/tech/drc`.
```suggestion
    result_dbs.append(run_check(rule_deck_full_path / "ihp-sg13g2.drc", table_name, layout_path, run_dir, switches))
```

ihp-sg13g2/libs.tech/klayout/tech/drc/rule_decks/beol/6_10_sealring.drc
dnltz: Address changed location here.
```suggestion
  layers_def_path = File.join(script_dir, '../layers_def.drc')
```

ihp-sg13g2/libs.tech/klayout/tech/drc/run_drc.py
mole99: Thanks for the catch! I've now tested the deck in an independent location:

```
python3 $PDK_ROOT/$PDK/libs.tech/klayout/tech/drc/run_drc.py --path=$PDK_ROOT/$PDK/libs.tech/klayout/tech/drc/testing/testcases/unit/activ.gds --run_mode=deep --run_dir=test_activ --no_density --mp 1
python3 $PDK_ROOT/$PDK/libs.tech/klayout/tech/drc/run_drc.py --path=$PDK_ROOT/$PDK/libs.tech/klayout/tech/drc/testing/testcases/unit/activ.gds --run_mode=deep --run_dir=test_activ --no_density --mp 16
python3 $PDK_ROOT/$PDK/libs.tech/klayout/tech/drc/run_drc.py --path=$PDK_ROOT/$PDK/libs.tech/klayout/tech/drc/testing/testcases/unit/activ.gds --run_mode=deep --run_dir=test_activ --no_density --table activ --mp 1
python3 $PDK_ROOT/$PDK/libs.tech/klayout/tech/drc/run_drc.py --path=$PDK_ROOT/$PDK/libs.tech/klayout/tech/drc/testing/testcases/unit/activ.gds --run_mode=deep --run_dir=test_activ --no_density --table activ --mp 16
```

ihp-sg13g2/libs.tech/klayout/tech/drc/rule_decks/beol/6_10_sealring.drc
mole99: Thanks! Fixed as well.

sergeiandreyev:
Hi @atorkmabrains, @FaragElsayed2, could you please help with review?..

mole99:
Rebased on top of latest `dev`, the CI is clean.

dnltz:
@atorkmabrains @FaragElsayed2 would be nice if you can give some feedback because my PR #731 depends on this one here.

atorkmabrains:
@dnltz and @mole99 Please add a new github workflow and run this new setup on a small GDS to be able to verify that it would function correctly.

The changes looks in generally correct, but I'm not sure how this would affect the interaction of the layers.

cc @sergeiandreyev

mole99:
@atorkmabrains Thank you for reviewing the PR.
Ideally, I would rather not store the layout of a larger design in the PDK itself, as this would bloat the size of the git repository.

However, this is an opportunity to swat two flies with one blow: I wanted to add LibreLane to the CI anyway.
This way LibreLane can implement a simple design and subsequently run KLayout DRC on it.

atorkmabrains:
@mole99 I'm not asking you add a large design. Just a extremely small GDS with may be one cell. Or you could just run it on one of the standard cells or constant PCells available already to see the logs in the run. I need to validate fundamentally that it will run without any basic syntax issues or layer duplications. I hope that make sense to you.

Please read my comments carefully.

cc @sergeiandreyev

dnltz:
@atorkmabrains can we not just provide a log here? I agree we should have a check for this script, but adding a CI pipeline only for seeing the logs seems like the wrong way (in my opinion).

@mole99 I would recommend to check LibreLane in a dedicated repository. That's basically why I maintain https://github.com/aesc-silicon/i2c-gpio-expander. It's running nightly checks for ORFS and this PDK to verify everything is still fine. We could also migrate that from ORFS to LibreLane? BTW: github actions might not have enough disk space to build simple chips ;)

atorkmabrains:
@dnltz No, it has to be some form of regression. It should be very quick to run on a single cell design. It will just tell me if there is any syntax issue or integration issues right away and it will make sure in the future that this change won't break with any updates. If you think my approach is wrong, could you please clarify why? And again, this is not a full regression, this is just a smoke test to make sure nothing breaks. It's not just viewing logs, you should check if klayout errors out or not and completes the run or not.

Again, you don't need any design, just pick any cell of the standard cells library in this GDS and run this rule deck on it:
https://github.com/IHP-GmbH/IHP-Open-PDK/blob/main/ihp-sg13g2/libs.ref/sg13g2_stdcell/gds/sg13g2_stdcell.gds

cc @sergeiandreyev

mole99:
@dnltz I agree, for larger designs the CI should run in a separate repository, which it does for the LibreLane template design: https://github.com/IHP-GmbH/ihp-sg13g2-librelane-template
However, it makes sense to run a small smoke test to ensure future changes to the LibreLane config or the DRC/LVS setup for either magic or KLayout do not break LibreLane.

@atorkmabrains With the latest commit, I have added a smoke test to the "DRC Regression Testing" for the top-level DRC deck. This ensures that it cannot break silently in the future. However, please bear in mind that `run_drc.py` already uses the top-level DRC deck, so any syntax errors would also be caught this way.

With this, all of the requests have been addressed.
@sergeiandreyev, the PR is ready to merge.

atorkmabrains:
@mole99 Great Job. And you are absolutely correct. It seems that you made changes to tables paths as well.

But I realized another thing that would need one last test similar to the one you added. Can you add another test on the same file with multi processing using run_drc.py to verify that it would work as well. You could use 2 as multiprocessing. This will verify that we are still able to split.

Thanks for the update.

Cc @sergeiandreyev

FaragElsayed2:
Hi @mole99 

I tried running the following command to test multiple table selections:

```bash
python3 run_drc.py --path=testing/testcases/unit/activ.gds --table=nwell --table=activ --table=cont --no_density --disable_extra_rules
```

It looks like only the first table is being processed and the others are ignored. Could you please take a look?

mole99:
@atorkmabrains I added another test for `run_drc.py` with multiprocessing.

@FaragElsayed2 Yes, multiple table selections are not supported. I wasn't aware that this feature was supposed to work, since in single-threaded mode it would always pick the first table name, if I see that right: https://github.com/IHP-GmbH/IHP-Open-PDK/blob/92974fb80c7d22ab6e6b5a859049f0b26e2c71e7/ihp-sg13g2/libs.tech/klayout/tech/drc/run_drc.py#L766

However, now that you have mentioned it, I'm thinking that we can make `table_name` an array and thus support multiple tables even for single threaded operations.
Of course, instead of e.g. `if TABLE_NAME == 'metal1' || BEOL`, it needs to be `TABLE_NAME.include?('metal1') || BEOL`.
I'll ping you once I've made the changes.

mole99:
@FaragElsayed2 Multiple table selections are now implemented.
It would be great if you could give this a try. I tested `run_drc` with and without mp, as well as the DRC deck directly.

mole99:
@FaragElsayed2 It would be great if you could take a quick look, so this PR can be merged.

atorkmabrains:
@mole99 Please add a test case with multiple selection in the regression just like you did for everything else.

Just to be clear, any feature you add you must add an equivalent regression "Smoke Test" to verify. 

I would recommend add a new section with multiple random combinations of tables (from 5 to 10) at least on any small design and verify it didn't generate any "Syntax errors". And you must do that both from direct klayout and run_drc.py. I check the drc regression if you have added those tests.

cc @sergeiandreyev

FaragElsayed2:
@mole99 

I just tested it, and everything looks great now. Thanks for your changes!

atorkmabrains:
@FaragElsayed2 did manual testing, but that is not enough for me.

@mole99 Please add the test cases that I have requested.

cc @sergeiandreyev

atorkmabrains:
@sergeiandreyev @mole99 @dnltz I don't recommend merging without the extra regression requested above.

mole99:
@FaragElsayed2 thank your for the manual testing, and for confirming that it works for you 👏

@atorkmabrains

> Just to be clear, any feature you add you must add an equivalent regression "Smoke Test" to verify.

What you are doing here feels a little pedantic, given that multiple tables in single-threaded mode were broken in `run_drc.py` and there was no test for this. That was the reason why I thought this was an unsupported/experimental feature.

I will add yet another smoke test for multiple tables, but I have no plans to do what you are requesting here:

> I would recommend add a new section with multiple random combinations of tables (from 5 to 10) at least on any small design and verify it didn't generate any "Syntax errors". And you must do that both from direct klayout and run_drc.py. I check the drc regression if you have added those tests.

This does not make sense. If you run the full DRC deck with all rule decks enabled and you don't get any syntax errors, you won't get syntax errors with a subset of tables selected.

@sergeiandreyev at this point, this is becoming absurd. I have demonstrated that these changes do not regress any of the supported features and do not introduce any "Syntax Errors".

atorkmabrains:
@mole99 You have already done the heavy lifting. And really thanks for the hard work so far. I believe the changes that I'm asking is very easy to implement give what you have done already. You could ask chatgpt even to write it for you.

And please trust me on this, I have been doing this for 19 years now.

And I can't say this "Unsupported/Undocumented", you have made changes in `run_drc.py` script. It's impacting the run methodology we provided.

I'm not trying to give you hard time as you might think. Actually, if my team was doing this, I would have asked for more regressions than what I have asked you todo. I'm really going easy on you from the requirements point of view. That's why I refused to do this task, because it would open a whole can of worms that I really want to avoid. @FaragElsayed2 could attest to that.

Please be patient and appreciate your cooperation.

@sergeiandreyev I leave the final judgement call with you as a leader for this project, but if you really depend on my experience my recommendation is to complete the request that I have asked.

atorkmabrains:
@mole99 

For example, one of the things that we didn't test is how this impacts the results. I didn't ask for that, which I would normally do with my team.

As I said, there many things I didn't ask you todo.

I'm just doing the basic smoke tests to make sure we have a smooth "Start" with the hope that this won't impact the actual results.

Appreciate your patience and understanding.

cc @sergeiandreyev

mole99:
@sergeiandreyev as mentioned before, I have now added smoke tests for runs with selected tables.

atorkmabrains:
@mole99 One of the runs gave the following error:
```
2025-11-12 12:01:18 +0000: Memory Usage (451816K) : Error reading default DRC rules from : no implicit conversion of nil into String
```

Please verify that this not related to the import of a file.

Please check:
https://github.com/IHP-GmbH/IHP-Open-PDK/actions/runs/19296545831/job/55179844151?pr=728#step:7:333
https://github.com/IHP-GmbH/IHP-Open-PDK/actions/runs/19296545831/job/55179844151?pr=728#step:8:334

cc @sergeiandreyev

FaragElsayed2:
Hi @mole99,  

I just wanted to clarify why this error occurs and what impact it has.  

We’re using a JSON file as input to the `run_drc.py` script (as requested by the IHP team). This JSON contains all DRC parameter values for every rule. It’s not mandatory for the user to provide one, since we have two fallback options:  

1. The main JSON file included in the repository:  
   `ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/sg13g2_tech_mod.json`  
2. A secondary backup file located alongside the rule decks:  
   `sg13g2_tech_default.json`  

Both of these files are automatically set in `run_drc.py`. If neither of them exists, the script will raise an error.  

However, when you run the full DRC deck directly from the KLayout command line like this:  

```bash
klayout -b -r ihp-sg13g2/libs.tech/klayout/tech/drc/ihp-sg13g2.drc \
    -rd run_mode=deep \
    -rd threads=1 \
    -rd no_feol=false \
    -rd no_beol=false \
    -rd no_offgrid=false \
    -rd no_pin=false \
    -rd no_forbidden=false \
    -rd input=ihp-sg13g2/libs.tech/klayout/tech/drc/testing/testcases/unit/activ.gds
```

and you don’t specify either drc_json or drc_json_default, then no parameter values are loaded. As a result, all rule parameters default to 0.0 µm, which effectively breaks all rule checks.

You can easily verify this by comparing the generated .lyrdb file from a direct KLayout run with the one produced when using `run_drc.py`.

atorkmabrains:
@mole99 Based on this feedback from @FaragElsayed2, I have to enforce the need to run regression using the new klayout interface and you have to verify that all regressions are passing the actual tests.

This has to be a totally separate run from the one we already have.

cc @sergeiandreyev

atorkmabrains:
@mole99 The last change you made will not fix the root cause, this will make it show. But as @FaragElsayed2 has mentioned all values are zero.

@mole99 You also need now to do a full switch checking on both variants of DRC rule decks. I can't accept the PR without this.

cc @sergeiandreyev

mole99:
@FaragElsayed2 Thanks for the explanation, but I have already found that `$drc_json_default` was missing a default value :)
I supplied the default value locally, but forgot to add it as an argument for the CI. You can see this in previous runs as well.

More importantly, the DRC deck should raise an error instead of continuing execution with empty DRC rules.

It seems that the previous code did just silently return from `get_drc_values`:

```
-    return {}
+    raise "Error reading default DRC rules from #{$drc_json_default}: #{e.message}"
```

This has now been fixed in the latest commit.

FaragElsayed2:
> @FaragElsayed2 Thanks for the explanation, but I have already found that `$drc_json_default` was missing a default value :) I supplied the default value locally, but forgot to add it as an argument for the CI. You can see this in previous runs as well.
> 
> More importantly, the DRC deck should raise an error instead of continuing execution with empty DRC rules.
> 
> It seems that the previous code did just silently return from `get_drc_values`:
> 

Hi @mole99,

This functionality is implemented in `run_drc.py`, which serves as our supported interface. We didn’t originally support running the KLayout rule deck directly — as you may recall, we didn’t even have the ihp-sg13g2.drc file at that time.

```bash
    # Locate JSON rule file paths
    script_dir = Path(__file__).resolve().parent
    tech_rule_path = (
        script_dir / "../../python/sg13g2_pycell_lib/sg13g2_tech_mod.json"
    ).resolve()
    default_rule_path = (script_dir / "rule_decks/sg13g2_tech_default.json").resolve()
    switches["drc_json_default"] = default_rule_path

    # Select best-available JSON rules file
    if arguments.drc_json:
        user_path = Path(arguments.drc_json).resolve()
        logging.info(f"User-provided DRC rules JSON: {user_path}")
        candidate_paths = [user_path, tech_rule_path, default_rule_path]
    else:
        candidate_paths = [tech_rule_path, default_rule_path]

    for path in candidate_paths:
        if path.is_file():
            logging.info(f"Using DRC JSON file: {path}")
            switches["drc_json"] = path
            break
    else:
        logging.error("No valid DRC JSON file found in any expected location.")
        raise FileNotFoundError("DRC JSON not found in user or fallback paths.")
```

Another note regarding this point:

`given that multiple tables in single-threaded mode were broken in run_drc.py`

I tested this in my dev (same dev branch as in this repo), and it seems to be working as expected.

Thanks a lot for all the updates you’ve made, really appreciate your efforts.

mole99:
Based on your feedback I've now also added `sg13g2_tech_mod.json` as the primary tech rule file.

mole99:
@FaragElsayed2

> This functionality is implemented in run_drc.py, which serves as our supported interface. We didn’t originally support running the KLayout rule deck directly — as you may recall, we didn’t even have the ihp-sg13g2.drc file at that time.

Yes, that is absolutely fine. I'm not blaming you at all for not assigning a default value - that wasn't a use case for you. However, the whole point of this PR is to make it possible to run the DRC Deck also directly.

> I tested this in my dev (same dev branch as in this repo), and it seems to be working as expected.

Huh, that surprises me. Given the code snippet I send earlier, `table_name` was assigned the first table name that is in `args.table`. If the first table does not require connectivity rules, but the second table does, surely this will cause issues?

> Thanks a lot for all the updates you’ve made, really appreciate your efforts.

Thanks also for the great work you've been doing!

atorkmabrains:
@mole99 Could you please add an extra DRC regression that would run using the Klayout only version?

cc @sergeiandreyev

atorkmabrains:
@mole99 As I mentioned, you need to do all switch/tables combination checking.

cc @sergeiandreyev

atorkmabrains:
@sergeiandreyev You can merge this PR, but keep in mind there is a potential it might break things and come back with issues related to certain combinations of tables.

mole99:
@sergeiandreyev On my end, this PR is ready to be merged.
The CI issue has been resolved and the DRC Regression Testing is successful.

If any issues related to the top-level DRC arise, I will definitely take a look at them. I'm personally interested in keeping the possibility of running the top-level DRC directly.

PR #731, opened 2025-10-28
DRC: Add `no_recommended` flag

dnltz:
Add a new command-line flag to disable recommended rules. Right now, this flag will disable some pad rules, which are always present in designs with IO ring.

- [ ] Based on #728

PR #733, opened 2025-10-29
Update Librelane Configuration and Add Pad Configuration

mole99:
This PR brings the LibreLane configuration up to date with the latest PDK version.

Additionally, the changes include a pad configuration under `librelane/sg13g2_io/`.
This pad configuration, in combination with the LibreLane pad ring generation, allows to implement full chip designs.

Currently, LibreLane runs KLayout DRC in a single process, as it was the case before. Support for the multiprocessing using `run_drc.py` will be added in a follow-up PR.

The liberty files for the I/O cells were missing the analog pad, therefore, I added it.
In order to automatically create the global connections, LibreLane requires the Verilog views of the power/ground pads to have power/ground ports. I added these as well.

- [x] I will open issue reports for these two issues in the [c4m-flexio](https://gitlab.com/Chips4Makers/c4m-flexio) repository.
  - https://gitlab.com/Chips4Makers/c4m-flexio/-/issues/7
  - https://gitlab.com/Chips4Makers/c4m-flexio/-/issues/8

Detailed changes:

- Use RC values from OpenROAD for better SPEF match
- Use sg13g2_buf_16 as clock root buffer
- Add sg13g2_io pad configuration
- Add power/ground ports for power/ground pads
- Add sg13g2_IOPadAnalog to all libs
- Update LibreLane config to latest PDK
- Update DIODE_CELL to include the port
- Update the seal ring script

Depends on #728
Enables https://github.com/librelane/librelane/pull/806

### DIFF
--- a/ihp-sg13g2/libs.tech/librelane/config.tcl
+++ b/ihp-sg13g2/libs.tech/librelane/config.tcl
@@ -59,6 +59,7 @@ dict set ::env(KLAYOUT_DRC_OPTIONS) beol true
 dict set ::env(KLAYOUT_DRC_OPTIONS) pin true
 dict set ::env(KLAYOUT_DRC_OPTIONS) forbidden true
 dict set ::env(KLAYOUT_DRC_OPTIONS) run_mode deep
+dict set ::env(KLAYOUT_DRC_OPTIONS) no_recommended true
 dict set ::env(KLAYOUT_DRC_OPTIONS) drc_json_default "$::env(PDK_ROOT)/$::env(PDK)/libs.tech/klayout/tech/drc/rule_decks/sg13g2_tech_default.json"
 
 set ::env(KLAYOUT_DENSITY_RUNSET) "$::env(PDK_ROOT)/$::env(PDK)/libs.tech/klayout/tech/drc/rule_decks/density.drc"


ihp-sg13g2/libs.ref/sg13g2_io/verilog/sg13g2_io.v
```diff
@@ -273,27 +273,47 @@ endmodule
 // type: IOVss
 `timescale 1ns/10ps
 `celldefine
-module sg13g2_IOPadIOVss ();
+module sg13g2_IOPadIOVss (
+    inout vss,
```
dnltz: I'm not quite sure about this change. Power and ground should be invisible at this level. Especially since the `iocss` and `iovdd` pins are missing for other IO cells and it's inconsistent now.

Why did you add those pins?

mole99: Please see this issue for a discussion about the power ports: https://gitlab.com/Chips4Makers/c4m-flexio/-/issues/7

I agree about the inconsistency, I can add the power ports to the other I/O pad cells as well.

The I/O cells in gf180mcu do not use `USE_POWER_PINS` ifdef guards around the power ports, while the sky130 I/O cells do. Do you by any chance have any experience how proprietary PDKs handle that?

dnltz: I would appreciate if those pins could be guarded by `USE_POWER_PINS` here too. This change will break all existing designs which don't need those pins.

mole99: Does it break those designs? I thought you can simply ignore them.

The new "open source" PDK also uses power/ground ports without guards: https://github.com/openecos-projects/icsprout55-pdk/blob/main/IP/IO/ICsprout_55LLULP1233_IO_251013/verilog/icsIOA_N55_3P3.v

dnltz: I did some tests and it still works when adding those pins to `IOPadIn`. I'm actually placing those power pins via OpenROAD. They're not part of my RTL.

ihp-sg13g2/libs.ref/sg13g2_io/verilog/sg13g2_io.v
```diff
@@ -273,27 +273,47 @@ endmodule
...
 // type: IOVdd
 `timescale 1ns/10ps
 `celldefine
-module sg13g2_IOPadIOVdd ();
+module sg13g2_IOPadIOVdd (
+    inout vss,
+    inout vdd,
+    inout iovss,
+    inout iovdd
+);
```
dnltz: Can you please stick to Verilog 1995? All other IO cells are defined in this older standard too.
```suggestion
module sg13g2_IOPadIOVdd (vss, vdd, iovss, iovdd);
    inout vss;
    inout vdd;
    inout iovss;
    inout iovdd;
```

ihp-sg13g2/libs.ref/sg13g2_io/verilog/sg13g2_io.v
mole99: Great! Yes, you can set the power connections using `add_global_connection` in OpenROAD, this is also possible in LibreLane. Specifying the power connections in the RTL is simply another method.

ihp-sg13g2/libs.ref/sg13g2_io/verilog/sg13g2_io.v
mole99: Good point! I've changed the port definitions to Verilog 1995 and added power/ground ports to the other I/O cells.

mole99:
@sergeiandreyev This PR is now also ready to be merged.

- The changes to `libs.ref/sg13g2_io/` have been discussed with Staf (see above).
- The seal ring has been updated to enable direct placement in an existing design. (It is backwards compatible.)

All other changes concern the LibreLane configuration in order to enable pad ring generation.

sergeiandreyev:
@b10346, could you please take a look at the proposed `sg13g2_io` changes?

b10346:
> @b10346, could you please take a look at the proposed `sg13g2_io` changes?

Changes look correct. I have just two comments:
1) power pins in verilog models: should we have two variants of verilog models (with and without power pins) or not? should we have variant with power pins for standard cells as well? Just to keep these question in mind for planning next IO update.
2) Is it necessary to add 2x2 delay "lutable" in analog pad Liberty model? What is a purpose to have it?

mole99:
Hi @b10346,

- I wouldn't maintain two variants of the file. If necessary, we can add a Verilog define to enable the power ports.
  However, the "open source" ICSprout PDK also has power ports on the I/O cells: https://github.com/openecos-projects/icsprout55-pdk/blob/main/IP/IO/ICsprout_55LLULP1233_IO_251013/verilog/icsIOA_N55_3P3.v
  It does not have power ports on the stdcells: https://github.com/openecos-projects/icsprout55-pdk/blob/main/IP/STD_cell/ics55_LLSC_H7C_V1p10C100/ics55_LLSC_H7CH/verilog/ics55_LLSC_H7CH.v

- The `delay_template_2x2` is used by `sg13g2_IOPadAnalog` for the `rise_transition` and `fall_transition`. If I remove the table, there will be an error. I can't tell you whether this table actually makes sense to have. As Staf mentioned, the analog pad is not finished: https://gitlab.com/Chips4Makers/c4m-flexio/-/issues/8

sergeiandreyev:
some additional comments on power ports in open source I/O Verilog models:
- sky130 has both model variants in one file separated by `ifdef USE_POWER_PINS`
- gf180mcu is providing only one variant w/ power pins

mole99:
@sergeiandreyev So it should be fine to go with power pins by default, since ICSprout and gf180mcu do that.
We can add a `USE_POWER_PINS` power guard later if needed for some reason.

sergeiandreyev:
@mole99, I would agree with the proposed change (power pins in I/O cells) **only** as a temporary workaround:
- currently IHP OpenPDK Verilog models for I/O, stdcell, SRAM do not have power pins defined, this makes them consistent
- proposed change for only I/O Verilog models breaks this consistency
- sky130 Verilog models could serve as an example of consistent approach: stdcells, I/O, SRAM Verilog models all use the same `ifdef USE_POWER_PINS` approach and both model variants are in one file
- gf18mcu is a bad example: stdcells have `ifdef USE_POWER_PINS`, while I/O and SRAM cells have only one model variant w/ power pins included
- the ICSprout PDK, in its present state and with the data available, cannot be considered a viable reference option

mole99:
@sergeiandreyev That sounds fine to me.
In that case, will the standard cell models then be updated to consider the power/ground ports? On sky130 VPWR/VGND need to be 1/0 for the primitive to function.

mole99:
I have rebased the PR on top of `dev` and added the `no_recommended` flag from #731.
@sergeiandreyev This PR is ready to be merged.

sergeiandreyev:
> @sergeiandreyev That sounds fine to me. In that case, will the standard cell models then be updated to consider the power/ground ports? On sky130 VPWR/VGND need to be 1/0 for the primitive to function.

@mole99, yes - we started internal discussion on this topic some time ago

PR #735, opened 2025-10-30
Minor updates for Klayout G2 DRC

### LINKED ISSUE #689: `Cnt.c` in SRAM cells
I'm having a lot of violations of `Cnt.c` rule on SRAM cells:
Image

but this is not expected since these are inside `DigiBnd.drw` and we have a custom rule:
Image

FaragElsayed2:
- Updated the `cnt.c` DRC rule to check only outside the digibnd layer.
- Added a new rule: `cnt.c.digibnd`.
- Modified the generated nBuLay condition to apply for Nwell ≥ 3 µm instead of > 3 µm.

Fixed #689


FaragElsayed2:
@sergeiandreyev 

Could you please check?

FYI, there are still `cnt.c` violations in the SRAM cells with a value of 0.02 µm. I believe these are actual violations.

sergeiandreyev:
Hi @b10346, could you please double check if that's indeed the case?
> FYI, there are still cnt.c violations in the SRAM cells with a value of 0.02 µm. I believe these are actual violations.

Issue #743, opened 2025-11-07
ngspice simulation with wrong destructive read on SRAM block


stafverhaegen-chipflow:
cc @holvo @noherbrferurtth @OpenVAF

I stumbled on a bug when doing ngspice simulation on a SRAM block. Attached is a test bench ([tb_dp.tar.gz](https://github.com/user-attachments/files/23421846/tb_dp.tar.gz)) with all the simulations I performed.
TL;DR is that read operation results in a destructive read for one 512x32 SRAM design simulated with KLU solver and gear method. Changing simulation setup or using the 512x8 results in a right read operation.  
For the simulations reported in this issue ngspice from the IIC-OSIC-TOOLS 2025.09 docker image was used, e.g. ngspice v45.2.

I have two SRAM designs one with 512 words of 8 bits and one with 512 words of 32 bits. Here is the layout of the top cells:

_dp512x8_
<img width="492" height="512" alt="Image" src="https://github.com/user-attachments/assets/b15316ff-88c5-4fbe-a916-df4b449d97a5" />

_dp512x32_
<img width="1161" height="409" alt="Image" src="https://github.com/user-attachments/assets/9eb23978-aa0a-4952-8a15-13ab09cd6bad" />

Here same layout but showing sub blocks:

_dp512x8_
<img width="528" height="526" alt="Image" src="https://github.com/user-attachments/assets/1cade1a1-7f67-43ba-a00d-b0bef54cbc74" />

_dp512x32_
<img width="1169" height="442" alt="Image" src="https://github.com/user-attachments/assets/fc28ec2a-e785-4dd5-99f7-222a2eb35877" />

Most of the design is the same between the two, main difference is that a column that contains 8 bits is repeated one time for 512x8 and four times for 512x32.

You can also open the included `tb_dp.oas` file to view the design. There it is also clear most of the design is shared between the two blocks. The layout is only given for reference and better understand the test bench; simulation discussed here are done on the pre layout SPICE netlist; thus without parasitics.

The test bench itself consists of a run of two clock cycles. In the first clock cycle $F(FFF) is written in address 0 on port 1 and 0 in address 511 on port2. In the second clock cycle the contents of address 511 is read on port 1 and the one of address 0 on port 2.

_dp512x8 port1_
In a first plot the action for the two clock cycles on port 1 is shown
[...]

PR #747, opened 2025-11-19
lv and hv mos: Split ngspice simulation library for cmos and rf-cmos applications

dwarning:
This PR is initiated by issue #743.
 
Actual all lv and hv mos simulations use by default the psp103_nqs model with switched off rf capabilities (rfmode=0). This is waste of resources and follows sometimes in convergence problems.

This PR allows the user to use the more suitable model psp103 for normal cmos designs in this 130nm process node. If more rf model features are required for rf circuits the user can switch to the psp103_nqs model by setting the device parameter rfmode=1 in the device symbol property box. There is no need to modify library entries.

Experiments with the 2k SRAM from issue #743 show runtime improvements of more than 50% with same results.


KrzysztofHerman (approved):
I ran the examples we have and it seems to work. 
It requires the `osdi` regeneration.

Issue #364, opened 2025-02-17
LVS feature request: option for implicitly connecting nets


stafverhaegen-chipflow:
In my development of the SRAM compiler I'm now doing LVS on subblocks. In these subblocks I can have more than `vdd` net that are not connected to each other. They each have their pin and are meant to be connected on a higher level.
klayout has a function `connect_implicit()` that allows to virtually connect the net. I would like to have an option to add nets to be connected virtually to the LVS deck. I'm using the LVS script, not the klayout GUI.
Other option is to have a `connect_implicit("*")` and have an option for switching between top-level and non top-level LVS.
Some context can be found in https://github.com/KLayout/klayout/issues/1986

stafverhaegen-chipflow:
cc @atorkmabrains @noherbrferurtth

atorkmabrains:
@sergeiandreyev

sergeiandreyev:
> [@sergeiandreyev](https://github.com/sergeiandreyev)

discussion in progress..

KrzysztofHerman:
Hi @stafverhaegen-chipflow, PRs #800, #828, and #862 have been merged. Do these changes resolve the issue? If yes, please consider closing this issue. Thanks!

stafverhaegen-chipflow:
Closing for now. Will file new issue if I am missing something for my SRAM generator.

PR #764, opened 2025-12-18
[WIP]Adapted LVS deck for SRAM generator of the FlowSpace project

stafverhaegen-chipflow:
In the FlowSpace project a SRAM generator is being developed. The SRAM blocks generated by the current version of the generator do not pass the normal LVS deck. This is because of two reasons:
* The layout uses overlapping transistor source/drain regions in between cells and this won't pass LVS without some specific hacking.
* There are some cells in the lowest part of the hierarchy where there disjointed nets for vss/vdd and nwell but are supposed to be the same net. Part of the reason is that in the lowest part of the hierarchy nwell and bulk may not have been connected and are thus floating.

The changes are consisting of smaller independent patches that each have their own description of what it does.
In [SP01.tar.gz](https://github.com/user-attachments/files/24240247/SP01.tar.gz) you find a top cell that won't pass the regular LVS but passes the adapted LVS.

I put forward this as a draft PR so further LVS development can be steered to take this into account. When the SRAM generator will become part of the PDK this will also need to be tackled.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/lvs/rule_decks/general_derivations.lvs
+++ b/ihp-sg13g2/libs.tech/klayout/tech/lvs/rule_decks/general_derivations.lvs
@@ -71,9 +71,11 @@ ngate_hv_base = ngate.and(thickgateox_drw)
 pgate_hv_base = pgate.and(thickgateox_drw)
 
 # S/D FETs
-nsd_fet = nactiv.not(nwell_drw).interacting(ngate).not(ngate).not_interacting(res_mk)
-psd_fet = pactiv.and(nwell_drw).interacting(pgate).not(pgate).not_interacting(res_mk)
+nsd_fet_propagated = nactiv.not(nwell_drw).interacting(ngate).not(ngate).not_interacting(res_mk)
+nsd_fet = nactiv.and(nsd_fet_propagated)
 
+psd_fet_propagated = pactiv.and(nwell_drw).interacting(pgate).not(pgate).not_interacting(res_mk)
+psd_fet = pactiv.and(psd_fet_propagated)
 
 # tap1 labels
 well_patt = glob_to_case_insensitive_glob("well")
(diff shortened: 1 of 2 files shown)


stafverhaegen-chipflow:
With this patch I don't need #364 anymore as the nets to implicitly connect are hard coded in the patch.

ihp-sg13g2/libs.tech/klayout/tech/lvs/rule_decks/general_derivations.lvs
```diff
@@ -71,9 +71,11 @@ ngate_hv_base = ngate.and(thickgateox_drw)
 pgate_hv_base = pgate.and(thickgateox_drw)
 
 # S/D FETs
-nsd_fet = nactiv.not(nwell_drw).interacting(ngate).not(ngate).not_interacting(res_mk)
-psd_fet = pactiv.and(nwell_drw).interacting(pgate).not(pgate).not_interacting(res_mk)
+nsd_fet_propagated = nactiv.not(nwell_drw).interacting(ngate).not(ngate).not_interacting(res_mk)
```
atorkmabrains, line 74: @stafverhaegen-chipflow What is the use of this layer?

stafverhaegen-chipflow:
> @stafverhaegen-chipflow and @sergeiandreyev could you please confirm what is the use of fet_propagated?

This came from Mattias in order to get the 'cheat' to work right:
https://github.com/KLayout/klayout/issues/1935#issuecomment-3429271884

KrzysztofHerman:
@stafverhaegen-chipflow could you please take a look at #862 PR, It should integrate (via. #828 ) your changes.

stafverhaegen-chipflow:
@KrzysztofHerman 
Will see if it works for the current tape-out of March. I will make new or follow-up PR if I am missing something. This may be after the tape-out deadline.

PR #776, opened 2026-01-10
libs.tech/librelane: use existing I/O sites, add site for corners

mole99:
This PR updates the LibreLane config to use the existing I/O site as well as a corner site. This eliminates the need to create fake sites.

There does not currently exist a corner site in the IO LEF, thus this PR adds `sg13g2_cornerSite`. An issue is documented in c4m-flexio: https://gitlab.com/Chips4Makers/c4m-flexio/-/issues/9


b10346:
The change is reasonable. It can be accepted.

PR #778, opened 2026-01-13
libs.tech/magic: add minimum and maximum capacitance corners for pex 

mole99:
This PR adds the minimum and maximum capacitance corners to the magic extraction setup.
The values were generated using [capiche](https://github.com/RTimothyEdwards/capiche).

Additionally, the min and max corner resistances for metal7 were adjusted, as otherwise the values would be ignored with:

> Resist argument must be integer or "None".

@RTimothyEdwards Could you please take a look at the changes?

### DIFF
--- a/ihp-sg13g2/libs.tech/magic/ihp-sg13g2-extract.tech
+++ b/ihp-sg13g2/libs.tech/magic/ihp-sg13g2-extract.tech
@@ -110,7 +110,8 @@ variants (hrhc),(hrlc)
  resist (allm4)/metal4           103
  resist (allm5,*mimcap)/metal5   103
  resist (allm6)/metal6            21
- resist (allm7)/metal7            14.5
+ # actually 14.5, but must be integer
+ resist (allm7)/metal7            15
 
  contact alldiffcont   22000
  contact pc            20000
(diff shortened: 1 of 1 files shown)


sergeiandreyev:
Hi @RTimothyEdwards,
> magic results could be compared against one from Calibre if IHP are willing to generate one

yes, we have this in progress (but using Klayout-PEX Magic-based runs), and we have some issues with reproducing the same extraction settings that we used in proprietary PDK to characterize stdcells (like 'decoupled', etc.)

RTimothyEdwards:
@sergeiandreyev :  I am unfamiliar with how the proprietary tool setups work for parasitic validation, so any information you can provide on how to better match the validation setup will be appreciated.

mole99:
Thanks for the review, Tim!

There is actually the [IHP_rcx_patterns.rules](https://github.com/IHP-GmbH/IHP-Open-PDK/blob/main/ihp-sg13g2/libs.tech/librelane/IHP_rcx_patterns.rules) for the nominal corner, which was generated for ORFS with, I assume, a proprietary extractor.

I can use that one for comparisons with the nominal corner of the open source rules. I'll open the PR for the min/nom/max rules as soon as this PR has been merged.

sergeiandreyev:
> @sergeiandreyev : I am unfamiliar with how the proprietary tool setups work for parasitic validation, so any information you can provide on how to better match the validation setup will be appreciated.

we're preparing a testcase.. I’ll send it to both you and Martin

mole99:
@sergeiandreyev Thanks! In the meantime, could you please merge this PR? As Tim mentioned, the numbers are reasonable.

PR #782, opened 2026-01-14
libs.tech/klayout: improve defaults in scripts

mole99:
This PR improves the defaults in the KLayout scripts.

- Improved defaults:
  - Now, `antenna.drc` and `density.drc` also set the default drc json files. This was already the case for `ihp-sg13g2.drc`.
  - `$threads` is set to the maximum number of CPU threads in all scripts if not specified explicitly.
  - `$run_mode` is set by default to 'deep'. While the DRC function `deep` was already called when no run mode was specified, `$run_mode` was left undefined. Thus, `$run_mode == 'deep' ? ps : ps.merged` would resort to merging the polygons in deep mode, which destroys the performance.

- Expand the file paths when loading the drc json files to improve readability.

- Do not buffer stdout: If one of the KLayout scripts gets stuck early on, it can happen that LibreLane shows no output, providing no hint as to what's the issue. This change ensures that the output is written to stdout.

Note that most of these issues do not arise if the arguments are explicitly set. However, these changes are useful for direct debugging.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/drc/ihp-sg13g2.drc
+++ b/ihp-sg13g2/libs.tech/klayout/tech/drc/ihp-sg13g2.drc
@@ -41,6 +41,9 @@ file_logger.formatter = formatter
 stdout_logger = Logger.new($stdout)
 stdout_logger.formatter = formatter
 
+# Don't buffer stdout
+$stdout.sync = true
+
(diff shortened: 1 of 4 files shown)


atorkmabrains (approved):
@sergeiandreyev Looks great to me.

sergeiandreyev (approved):

PR #790, opened 2026-01-23
fix(klayout/drc/dialogs): remove enforced background color

tskaar:
##  What
In the case where one uses a darker Qt theme, enforcing white background color, while not enforcing text color causes the text and background color to potentially be the same color.

Avoid enforcing background color, and let the Qt theme handle it.
The other enforced colors are looking _OK_ with my current theme, so I'll leave the changes to a minimum.

### DRC options

**Darker Theme**

| Before | After |
| --------|-------|
| image |  image |

**White/No Theme**

| Before | After |
| --------|-------|
|image |  image |

### DRC output logs

**Darker theme**

| Before | After |
| --------|-------|
| image |  image |

**White/No Theme**

| Before | After |
| --------|-------|
| image |   image | 

## Summary

For the default/no theme option, the only _functional_ change is that the title of the DRC output logs changes from a dark blue-grey (#2C3E50) to a straight dark color, however that impact is minimum I'd say.

Otherwise it should now work more consistently with the system-level Qt theme in terms of background color vs text-color

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/macros/sg13g2_drc.lym
+++ b/ihp-sg13g2/libs.tech/klayout/tech/macros/sg13g2_drc.lym
@@ -213,7 +213,7 @@ end
 @title_font.setPointSize(14)
 @title_font.setBold(true)
 @title_label.setFont(@title_font)
-@title_label.setStyleSheet("color: #2c3e50; margin-left: 10px;")
+@title_label.setStyleSheet("margin-left: 10px;")
 @header_layout.addWidget(@title_label)
 @header_layout.addStretch
 @main_layout.addLayout(@header_layout)
@@ -221,7 +221,7 @@ end
 # Log output area
 @log_output = RBA::QTextEdit.new
 @log_output.setReadOnly(true)
-@log_output.setStyleSheet("background-color: #f4f4f4; font-family: monospace;")
+@log_output.setStyleSheet("font-family: monospace;")
 @main_layout.addWidget(@log_output)
 
 # Close button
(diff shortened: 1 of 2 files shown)


sergeiandreyev:
I cannot test it since I don't use dark QT theme, maybe @KrzysztofHerman?

sergeiandreyev:
@atorkmabrains, are you ok w/ this PR?

tskaar:
> I cannot test it since I don't use dark QT theme...

If you have access/are able to install `adwaita-qt` you can (temporarily) set the theme using:

```
QT_STYLE_OVERRIDE=adwaita-dark klayout
```

From: https://github.com/KLayout/klayout/issues/2124

---

Alternatively if one is using KDE/Plasma one can set the (global) dark theme in KDE config and then run klayout using:

```
QT_QPA_PLATFORMTHEME=kde klayout
```
---

EDIT:

You can also do it more manually, by making a file, e.g: `dark.qss` with the content:

```css
QWidget {
    background-color: #1e1e1e;
    color: #e0e0e0;
}
```

And then launch KLayout with:

```
klayout -stylesheet=<path_to_dark.qss> 
```

Then all of the menus etc (with the exception of the layout background) will have a darker color, and the difference with this PR in the DRC windows should be visible.

sergeiandreyev:
looks nice, the text is visible, however some fields do not 'expand' correctly (in my case):
image

tskaar:
> looks nice, the text is visible, however some fields do not 'expand' correctly (in my case)

Ahh, I got the same when using adwaita-qt, apparently it's an upstream issue with that particular theme, .e.g https://github.com/FedoraQt/adwaita-qt/issues/200

I got the same issue whether I used the adwaita-qt (white) or adwaita-qt (dark), so it's a theming issue and not colorscheme issue :)

sergeiandreyev (approved):

Issue #815, opened 2026-02-11
Klayout(DRC Options): can trigger a segfault if opened multiple times


tskaar:
Hi!

When opening up the DRC options window in KLayout I noticed that it sometimes randomly crashes.
After some trial and error I realized that it's due to a "use-after-free" issue in the Ruby-Qt bindings, in particularly when using `QSpacerItem`.

## Workaround

Removing the `QSpacerItem` reference works around the issue, we can do that by applying the following patch:

```diff
diff --git a/ihp-sg13g2/libs.tech/klayout/tech/macros/sg13g2_drc_options.lym b/ihp-sg13g2/libs.tech/klayout/tech/macros/sg13g2_drc_options.lym
index a4d59f33..f033f627 100644
--- a/ihp-sg13g2/libs.tech/klayout/tech/macros/sg13g2_drc_options.lym
+++ b/ihp-sg13g2/libs.tech/klayout/tech/macros/sg13g2_drc_options.lym
@@ -178,7 +178,6 @@
     @main_layout.addLayout(@header_layout)

     ## --- Basic Settings Tab ---
-    @spacer = RBA::QSpacerItem.new(0, 5)
     @basic_tab = RBA::QWidget.new
     @basic_layout_setup = RBA::QFormLayout.new(@basic_tab)
     @basic_layout_setup.setSpacing(15)
@@ -186,7 +185,6 @@
     @basic_layout_setup.addRow("Run Mode:", create_combo_box(options['run_mode'], ['deep', 'flat']))
     @basic_layout_setup.addRow("DRC Table:", create_combo_box(options['table'] || "main", get_table_files))
     @basic_layout_setup.addRow("Parallel Runs:", create_cores_threads_no(options['parallel_runs']))
-    @basic_layout_setup.addItem(@spacer)
     @basic_layout_setup.setSpacing(10)
     # Checkboxes - BASIC tab
     chk_feol = create_check_box("Disable FEOL Checks", options['no_feol'], "Skips Front-End-Of-Line rule checks (e.g., active, poly).")
```

## Example

Here is a re-production done on my end

https://github.com/user-attachments/assets/9b0feb8e-9820-45f8-893f-c92d809a86a6

Note that reproduction is timing-dependent. You might need to open/close the instance rapidly to trigger the race condition, as shown in the attached example.

## Software

Klayout: v0.30.5
Qt: v6.10.2
Ruby: v3.4.8 +PRISM

## TODOs

[...]

PR #816, opened 2026-02-11
fix(KLayout/DRC options): explicitly unmanage `QSpacerItem` to avoid use-after-free segfault

tskaar:
Currently there is an issue with a potential use-after-free bug triggered in the Ruby-Qt bindings.
Explicitly call `unmanage` on the object to transfer the lifetime to C++ to avoid a potential segfault.

Fixes: #815


tskaar:
Re-factored based on feedback from Matthias [1].

\[1\]: https://github.com/KLayout/klayout/issues/2273#issuecomment-3886082845

PR #820, opened 2026-02-16
Add CLI options

hpretl:
For a smooth install in IIC-OSIC-TOOLS, the following CLI options have been added to the given scripts. The default behavior is NOT changed, so there is no breaking change for existing users.

- Refactor `ihp-sg13g2/libs.tech/qucs-s/install.py`:
  - Add argparse-driven CLI flags (`--no-model-compile`, `--no-qucs-check`, mutually exclusive `--no-qucs-workspace` / `--no-qucs-dir`).
  - Improve program checks, safer subprocess usage, rename/clean up helper function parameters, fix symlink/copy logic and messages, and make logging/messages more consistent.
- Update `ihp-sg13g2/libs.tech/verilog-a/openvaf-compile-va.sh`: 
  - Add support for a `--compile-model-generic` argument that sets `--target_cpu generic` when invoking the compiler
  - Update copyright/authors.

### DIFF
--- a/ihp-sg13g2/libs.tech/qucs-s/install.py
+++ b/ihp-sg13g2/libs.tech/qucs-s/install.py
@@ -2,14 +2,14 @@
 
 ########################################################################
 #
-# Copyright 2024 IHP PDK Authors
-# 
+# Copyright 2024-2026 IHP PDK Authors
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# 
+#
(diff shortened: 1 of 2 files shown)


hpretl:
@KrzysztofHerman Please look at this PR, as there is no impact to other users when the new CLI options are not used, a merge should be fine.

KrzysztofHerman:
@hpretl I will merge it. We are working on some global installation script so at some moment we will remove these  `*.py` and `*.sh` scripts and replace it by some global one.

hpretl:
@KrzysztofHerman Thanks for merging! It would be great to provide similar CLI options on the global script.

Issue #827, opened 2026-02-17
xschem integration : mkdir commands in tcl scripts in ihp-sg13g2/libs.tech/xschem directory


dgalayko:
In all below files/line number in the directory IHP-Open-PDK/ihp-sg13g2/[libs.tech/xschem](http://libs.tech/xschem), the line 
mkdir -p $netlist_dir
does not work in TCL. Should be : 
file mkdir $netlist_dir

Without that, examples provided with PDK don't work (see screenshot example). 

./xschem-menu:239:   
./start_page.sch:99:
./sg13g2_tests_xyce/dc_hbt_13g2.sch:152:
./sg13g2_tests/iso_dc_lv_nmos.sch:126:
./sg13g2_tests/dc_hv_nmos.sch:115:
./sg13g2_tests/iso_dc_hv_nmos.sch:116:
./sg13g2_tests/dc_hbt_13g2v_5t.sch:157:
./sg13g2_tests/dc_lv_nmos.sch:115:
./sg13g2_tests/dc_hbt_13g2l_5t.sch:159:
./sg13g2_tests/dc_hv_pmos.sch:118:
./sg13g2_tests/dc_lv_pmos.sch:119:
./sg13g2_tests/dc_hbt_13g2_5t.sch:157:
./sg13g2_tests/inv_sweep_tb.sch:70:
./sg13g2_tests/inv_sweep_tb.sch:252:
./sg13g2_tests/dc_hbt_13g2.sch:114:
./sg13g2_tests/inv_mc_tb.sch:61:
./sg13g2_tests/inv_mc_tb.sch:218:
./sg13g2_tests/isolbox_sweep_tb.sch:49:
./sg13g2_tests/isolbox_sweep_tb.sch:116:

dgalayko:
<img width="407" height="220" alt="Image" src="https://github.com/user-attachments/assets/e8c1c0a8-722f-47a9-9743-b9f35e223ca6" />

tskaar:
Related: https://github.com/IHP-GmbH/IHP-Open-PDK/pull/829
Related: https://github.com/IHP-GmbH/IHP-Open-PDK/discussions/699

sergeiandreyev:
Hi @dgalayko, could you please confirm that with the merged #829 the issue can be closed as fixed?

dgalayko:
Hello @sergeiandreyev, thank you very much for this fix, I tested the whole, everything looks fine. The issue can be closed

PR #829, opened 2026-02-18
Update dc_hv_nmos.sch

brunny95:
Missing exec gives an error. This happens for every testbench. Should update all of them.


tskaar:
Does it make sense to:

1) Fixup all the files in one shot?
2) Replace `exec` with `file mkdir` (as proposed in #827 ), since this is the more Tcl-idiomatic solution\[1\]?
    - `exec mkdir -p` -> `file mkdir`

\[1\]: https://www.tcl-lang.org/man/tcl8.6/TclCmd/file.htm#M22

KrzysztofHerman (approved):

KrzysztofHerman:
@brunny95 how about fixing all the files?

brunny95:
Done

dgalayko:
Hello, thank you very much for this fix, I tested the whole, everything looks fine.

PR #819, opened 2026-02-15
Update DRC rules and testcases to resolve reported issues

FaragElsayed2:
## DRC Issues

1. Update **AFil.d** Rule (Fixes #730)
   - Updated the rule to check space/enclosure against the original nBulay layer only, to match what we have in commercial PDK.
   - Updated the testcase to cover the enclosure part of the rule to pass regression.
   
 ----
 
 2. Update **Cnt.c** Rule (Fixes #689 and part of #793)

   - Updated the rule to differentiate enclosure checks based on region markers (sram, digibnd, and general) to match the commercial PDK.
  - Implemented specific values: 0.006µm for SRAM blocks (from commercial PDK as this rule existed in older DRM versions but is missing in the current one), 0.05µm for Digital Boundary (outside SRAM), and 0.07µm for the remaining areas.
  
 ----

3. Update **Sdiod.d/e** Rules (Fixes #684 and #793)

  - Updated sdiod.d and sdiod.e rules to be applied for Schottky devices only.
  - Confirmed that this update makes PDK cells, such as IO cells and SRAM cells, to pass DRC.

----

4. Fix NBL.c/d & NW.b1 Rules (Fixes #696)

  - Corrected the implementation logic for NBL.c, NBL.d, and NW.b1 rules.
  - This fix resolved false positive errors affecting Schottky devices.

----

5. Waive Svaricap DRC Violations (Fixes #639)
  - Waived Cnt.c, Cnt.e, Cnt.g1, Cnt.g2, and NBL.f violations for the svaricap device.
  - Confirmed that the svaricap cell is now DRC clean.
  
----

6. Update Offgrid Rules (Enables detection for #683)
  - Moved off-grid checks to a separate, clean rule deck to ensure coverage for all layers not excluded by the DRM.
  - Removed incomplete off-grid checks from the residual (maximum) rule deck.
  - 5 nm drawing grid on all relevant layers, explicitly excluding: DigiBnd, DigiSub, dfpad, EdgeSeal, HeatRes, HeatTrans, IND, NoDRC, NoMetFiller, NoRCX, RadHard, Recog, RES, Scribe, SRAM, and TEXT.

----

7. Add Cells DRC Regression (libs.ref)

Added a new regression script + CI workflow to run SG13G2 DRC on all reference cells from `libs.ref` (io / pr / stdcell / sram).

**Summary**

- **Total cells:** 179  
- **Excluded:** 1  
- **Cells with waived violations:** 26  

[...]

PR #837, opened 2026-02-24
drc: Fix RECOMMENDED and optimize FEOL

dnltz:
Fix how RECOMMENDED is handled in the `6_9_pad.drc` deck.

Also don't assign `nwell_nets = nwell_drw.nets` and rather use `nwell_drw.nets` directly in tests. This assignment doesn't respect `--no_feol` and takes forever on larger designs.


sergeiandreyev:
Hi @FaragElsayed2, could you please take a look?

ihp-sg13g2/libs.tech/klayout/tech/drc/rule_decks/beol/6_9_pad.drc
```diff
@@ -16,7 +16,7 @@
 # limitations under the License.
 #=========================================================================================
 
-if TABLES.include?('pad') || (BEOL && RECOMMENDED)
+if ((TABLES.include?('pad') || BEOL) && RECOMMENDED)
```
FaragElsayed2, line 19: Please check #819 — I have addressed it as part of one of the fixes in this PR.

In this file, only pad.fR is a recommended rule; the other rules are not recommended.

ihp-sg13g2/libs.tech/klayout/tech/drc/ihp-sg13g2.drc
```diff
@@ -376,9 +376,6 @@ if CONNECTIVITY_RULES
   connect(topvia1_n_cap, topmetal1_con)
   connect(topmetal1_con, topvia2_drw)
   connect(topvia2_drw, topmetal2_con)
-
```
FaragElsayed2, line 379: For optimization, we perform the derivation once and reuse it in the required rules. The connectivity section runs if we execute nwell, nbulay, or the full rule deck. Removing this line will not improve runtime, since the main processing time is spent in the connectivity stack.

As an optimization, I can guard the full connectivity section with the FEOL option so that it does not run when FEOL is disabled.

I have supported that now in #819.

FaragElsayed2: This can be updated later if any additional BEOL rules require connectivity.

dnltz: Ah perfect. The problem I have this that assignment takes hours with by 2500x2500um chip. I will open an issue soon. My solution right now is to disable FEOL entirely.

FaragElsayed2: @dnltz The runtime is not spent on this assignment itself; it is spent in the connectivity stack at the lines shown below.
image

We could consider adding an additional option to disable connectivity rules, as they significantly increase runtime for larger designs.

@atorkmabrains What do you think?

FaragElsayed2: @atorkmabrains We implemented something similar for gf180mcu, where we added a restricted value for those rules when the switch is disabled.

dnltz: @FaragElsayed2 I added `logger.debug` after each `connect`. They either connect in the background or it's the assignment, but it hang at the nwell assignment. Nevertheless, +1 for connecting only what's required.

FaragElsayed2: @dnltz 

That’s strange. From my experience, the main runtime cost comes from building connectivity, especially in flat mode, rather than the derivation itself.

Relocating the common assignment into individual rules isn’t ideal either, as it would require re-deriving shared layers across multiple rules/tables.

I understand the idea of skipping FEOL entirely, but I think introducing an additional switch would be a cleaner solution.

I’ll discuss this internally and work on addressing this KLayout optimization bottleneck, as we’ve encountered a similar issue before.

dnltz:
Closing this PR because #819 will fix it

ihp-sg13g2/libs.tech/klayout/tech/drc/ihp-sg13g2.drc
dnltz: I can send you my GDS file, if you want something for testing. Not urgent for the upcoming tape-out since I have a workaround.

PR #838, opened 2026-02-24
ihp-sg13g2: libs.ref: sg13g2_tech.lef: Fix TM2.bR

dnltz:
OpenROAD currently doesn't respect rules TM2.bR

"Min. space of TopMetal2 lines if, at least one line is wider than 5.0 µm and the parallel run is more than 50.0 µm"

Add a SPACINGTABLE to prevent TM2.bR violations.


dnltz:
FYI @b10346

b10346:
The change has to be accepted. Implemented and checked with internal flow.

PR #828, opened 2026-02-18
Update LVS scripts and tests to resolve reported issues - PR1

FaragElsayed2:
## LVS Fixes & Updates

1. Fixing LVS Net-Only Flow and Tap Behavior (Fixes #773 & #772)
    - Fixed --net_only so extraction options are applied correctly.
    - Improved LVS logging for options and comparison mode.
    - Aligned tap extraction to PDK order (TIE WELL) with explicit terminal mapping.
    - Kept taps as resistor-class for clearer LVS DB/debug view.
    - Enabled safe tap parallel combination while preserving terminal polarity.

image

----

2. Run scripts enhancement (Fixes #771)
  - Replaced docopt with argparse in LVS run/regression scripts
  - Improved --net_only behavior (including no-netlist handling and clear warnings)
  - Enhanced run_lvs.py logging with a clearer end-of-run summary table.
 
image

image

image

----

3. Fixing Caps issues (Fixes #418 & #459 & #494 & #512 & #555 & #796)
    - Some updates are based on work from #513
    - Resolved capacitor combiner and substrate-connection handling for MIM/RFCMIM devices.
    - Added  RFMIM cap testcase coverage for combiner scenarios (merge and no-merge cases).
image

    - Updated resistor series/parallel combiner behavior to avoid invalid merges and ensure parameter updates are applied consistently.
    - Note: The test provided in #555 currently has an open net issue; once it’s fixed, it will pass with the new updates.
    
image

----

4. Tested #347 
   - This issue was fixed in one of merged PRs, and it works as expected now 
image

----

5. Fixing #363 & #364  
    - Updating custom reader for better handling
    - Adding support for implicit net connections as comma separated list (CLI/GUI)

---

6. Enhancement for the LVS GUI
   - Using similar interface to what we have in DRC and categorize options 
   - Supporting the GUI log window for better visualization

image

image

---

7. Enhancement for svaricap (Fixes #753 & #414)
    - Add `Nx` parameter support for `sg13_hv_svaricap`.
    - Update svaricap unit testcase to cover `Nx=1..10` for both allowed size combos (small/large).
    - Refactor `DeviceCustomVaractor` to use explicit terminals `G1/W/G2/SUB`, and align varicap terminal mapping in  reader.
    - Enhance `GeneralNTerminalExtractor` flat-mode behavior by counting ports per `dev_mk` region.
image

--- 

8. Updates for ESD & Ptap device (Fixes #707)
    - Enable the use of ptap1 in ESD recognition layer.
[...]

PR #800, opened 2026-01-29
Proposal to Align the Xschem Netlist Format with KLayout LVS

H-Ojiro:
Fixes #<issue_number_goes_here>
**To align the Xschem netlist format with KLayout LVS, revised xschem sym files in ihp-sg13g2/libs.tech/xschem/sg13g2_pr**

1. Add new symbols: inductor.sym, inductor3.sym
image

2. Result of Klayout LVS using unit device cells

image

Note:
The CDL format for the resistor devices is revised as in the ngspice style;
```
RXXXXXXX n+ n- <value> <mname> <l=length> <w=width> ...
```
At present with this format, Klayout does not recognise the model name,  the resistor devices do not match

3. Here is a spreadsheet comparing the device statements writen by the Xschem Netlist;
https://docs.google.com/spreadsheets/d/1GEoOd517_JA-xxpHtrVB9OOGxybPtv_bV-c-KIgPGYg/edit?usp=drive_link
Refer to 'Netlist-Format' sheet.
Also images of the device  symbol are shown in 'Subcircuit-Symbols' sheet.

- [x] Tests pass
- [x] Appropriate changes to ~~README~~(Linked spreadsheets) are included in PR

PR #846, opened 2026-03-04
Dev->Main

KrzysztofHerman:
Merge request to update PDK before March, 30-th tape-out.

### DIFF
--- a/ihp-sg13g2/libs.tech/netgen/ihp-sg13g2_setup.tcl
+++ b/ihp-sg13g2/libs.tech/netgen/ihp-sg13g2_setup.tcl
@@ -270,20 +270,24 @@ lappend devices cap_rfcmim
 
 foreach dev $devices {
     if {[lsearch $cells1 $dev] >= 0} {
+	property "-circuit1 $dev" derive area w l
+	property "-circuit1 $dev" derive perimeter w l
 	property "-circuit1 $dev" parallel enable
-	property "-circuit1 $dev" parallel {l critical}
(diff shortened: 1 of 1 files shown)


ihp-sg13g2/libs.tech/xschem/sg13g2_pr/rhigh.sym
d-m-bailey, line 1: @KrzysztofHerman Same issues with `rhigh`, `rppd`,  and `rsil`.

1. I don't think this will propagate the connectivity as expected.

The `propag` value determines what pin will be shorted when probed with `shift-k`. Pins are numbered from `0`.
Since pin `P` is now the first pin, a `propag` property of `0` will do nothing. Similarly, a `propag` property of `1` on the second pin `M` will do nothing. The first pin should propagate to `1` and the second should propagate to `0`.

See the [xschem manual](http://repo.hu/projects/xschem/xschem_man/xschem_properties.html).

2. This change drops the `body` connection for LVS. Is that intended?

Note: These issues are fixed in #800

KrzysztofHerman:
@d-m-bailey   thank you for mentioning this. I will hold it on and review #800

dnltz:
@KrzysztofHerman I think #838 should be part of this merge.

#819 would be nice, but I can also add that to my local fork.

simi1505:
@KrzysztofHerman If somehow possible in time and the draft is finished, please also https://github.com/IHP-GmbH/IHP-Open-PDK/pull/828. We need a somewhat working LVS for the tapeout. Thanks a lot.

Issue #803, opened 2026-01-30
Extraction(cmim): bottom plate (Metal5)  doesn't act as a parasitic shield for top plate


tskaar:
Hi!

When using the `cmim` PCell, and one routes underneath (say using Metal4) it, I would expect that the main (parasitic) contributor would be the bottom plate wiring, however during extraction I get that the largest contributor is from the top-plate (TopMetal1).

Hence it doesn't seem like Metal5 (bottom plate of MIM) is acting as a shield (or is it completely ignored?) in this case, and it is TopMetal2 that acts as the "parasitic metal". 

The (potential) issue here is that `BOT` might be connected to ground, hence we can handle the additional capacitance, however `TOP` might be part of the signal chain, which is much more susceptible to parasitic capacitance

## Example

<img width="961" height="639" alt="Image" src="https://github.com/user-attachments/assets/73039604-1a19-46f6-a856-27e2fa8b68c3" />

**Left:**
By instantiating the `cmim` PCell with `w=50u l=20u`, and then putting the `TOP` pin on TopMetal1 left-edge boundary, and `BOT` pin on the Metal5 right-edge boundary, and then running a metal4 wire across it, with the pin `BUS` at the top.

**Right:**
In this case we simply flatten the PCell and leave _only_ Metal5 and TopMetal1 remaining, hence the `BUS_NC` in this case runs underneath to sheets of metal only, no components (hence the NC annotation).

### Extraction

When extracting this with Magic I get the netlist:

```
* NGSPICE file created from pex_cmim_shield.ext - technology: ihp-sg13g2

.subckt pex_cmim_shield TOP TOP_NC BOT_NC BOT BUS BUS_NC
X0 TOP BOT cap_cmim l=20u w=50u
C0 BOT_NC TOP_NC 44.0082f
C1 BOT_NC BUS_NC 4.71061f
C2 TOP BOT 1.97546f
C3 BOT BUS 0.44246f
C4 TOP BUS 4.38075f
C5 BUS_NC TOP_NC 0.11261f
C6 TOP_NC sub 4.62329f
C7 BOT_NC sub 11.3694f
C8 TOP sub 1.45041p
C9 BOT sub 4.5188f
C10 BUS_NC sub 4.78197f
C11 BUS sub 4.78197f
.ends
```

Where `BOT:BUS` and `TOP:BUS` for both with and without (`*_NC`) the PCell are listed.

### Sanity check - Fastercap

[...]

Issue #802, opened 2026-01-30
Extraction(cmim): double capacitance count for top-plate capacitor


tskaar:
Hi,

When extracting a MIM-capacitor I am essentially seeing a duplicate capacitor on the top-plate of the capacitor, between top-plate and substrate. 

Under the assumption that I have a net called TOP on the top-plate, and BOT on the bottom plate, i.e in layout it looks like this:

<img width="800" height="358" alt="Image" src="https://github.com/user-attachments/assets/85072b08-d04c-4f13-93f3-536f8213b2d9" />

Which is just a CMIM Pcell with `l=20u w=50u` and then the pins on the edge of TopMetal1 and Metal5 for TOP and BOT, respectively

When extracted, I get the following:

```
* NGSPICE file created from pex_cmim.ext - technology: ihp-sg13g2

.subckt pex_cmim TOP BOT
X0 TOP BOT cap_cmim l=20u w=50u
C0 TOP BOT 1.99968f
C1 TOP sub 1.51048p
C2 BOT sub 4.59561f
.ends
```

Note that I am getting 1.5 pF between TOP and sub.
Under nominal conditions the MIMcap has 1.5 fF/µm², hence we expect it to give us ~1.5 pF of capacitance with the numbers used, which is _suspiciously_ close to the numbers I got.

## Workaround

If I apply the following patch:

```patch
diff --git a/ihp-sg13g2/libs.tech/magic/ihp-sg13g2-extract.tech b/ihp-sg13g2/libs.tech/magic/ihp-sg13g2-extract.tech
index 5e223dda..faab0818 100644
--- a/ihp-sg13g2/libs.tech/magic/ihp-sg13g2-extract.tech
+++ b/ihp-sg13g2/libs.tech/magic/ihp-sg13g2-extract.tech
@@ -1121,8 +1121,8 @@ variants *
 
  # Area coefficient of the MiM cap for estimation of values for CDL
  # netlist output.
- defaultareacap     *mimcap metal5 1500
- defaultperimeter   *mimcap metal5 40
+ # defaultareacap     *mimcap metal5 1500
+ # defaultperimeter   *mimcap metal5 40
 
 # Devices:  Base models (not subcircuit wrappers)
```

And re-run the extraction I get:

```
* NGSPICE file created from pex_cmim.ext - technology: ihp-sg13g2

.subckt pex_cmim TOP BOT
X0 TOP BOT cap_cmim l=20u w=50u
[...]

PR #853, opened 2026-03-05
Corrects the MiM cap layers for parasitic extraction.

RTimothyEdwards:
Corrects the MiM cap layers for parasitic extraction. Previously the MiM cap layer was placed on metal5, but since the MiM cap layer is considered to be the MiM cap top plate and therefore connected to the metal layer above it, this arrangement was causing all parasitic capacitance to the bottom plate under the MiM cap layer to be added to the node of the top plate instead, which then ended up as an underestimate of parasitics to the bottom plate and an overesitmate of the parasitics to the top plate.  The solution is to create a separate plane for the MiM cap layer, not to put it on metal5.  This simplifies a number of DRC and extraction rules, which no longer need to include the mimcap layer as part of metal5.

This solves issue #803 "Extraction(cmim): bottom plate (Metal5) doesn't act as a parasitic shield for top plate".

Also:  Corrected the lines in the extraction section which allow the LVS extraction to compute a value for MiM caps, which were supposed to *only* be used for LVS extraction.  The lines of code were under "variants *" when they should have been under "variants (lvs)".

This solves issue #802 "Extraction(cmim): double capacitance count for top-plate capacitor".

Signed-off-by: R. Timothy Edwards <tim@opencircuitdesign.com>

- [ ] Tests pass

### DIFF
--- a/ihp-sg13g2/libs.tech/magic/ihp-sg13g2-drc.tech
+++ b/ihp-sg13g2/libs.tech/magic/ihp-sg13g2-drc.tech
@@ -608,34 +608,34 @@ variants *
 # METAL5 -
 #--------------------------------------------------
 
- width allm5,*mimcap,sealv5 200 "Metal5 width < %d (M5.a)"
- spacing allm5,*mimcap,sealv5  allm5,*mimcap,obsm5,sealv5 210 touching_ok \
+ width allm5,sealv5 200 "Metal5 width < %d (M5.a)"
+ spacing allm5,sealv5  allm5,*obsm5,sealv5 210 touching_ok \
 	"Metal5 spacing < %d (M5.b)"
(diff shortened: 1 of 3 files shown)


tskaar:
Can confirm that this fixes my issues (#802 , #803)

For "Extraction(cmim): double capacitance count for top-plate capacitor", I now get:

| Net     | Capacitance [Current] | Capacitance [this PR] |
|:-------:|:---------------------:|:---------------------:|
| TOP:BOT | 1.9754 fF             | 1.9754 fF             |
| TOP:BUS | **4.3807 fF**             | 0.1126 fF             |
| BOT:BUS | 0.4424 fF             | **4.7106 fF**             |

Which tracks the expectation.

And for "Extraction(cmim): bottom plate (Metal5) doesn't act as a parasitic shield for top plate", the double counting is gone

Thanks for the quick fix! :raised_hands:

PR #869, opened 2026-03-11
This quick fix removes a line from the magic startup script that is breaking LibreLane.

### LINKED ISSUE #868: Command in magic startup script breaks LibreLane
Unfortunately I left the command `units microns` in the magic startup script (`ihp-sg13cmos5l.magicrc`), which I was using to find scripts that are not compatible with the new unified method for handling dimensional units in magic.  If the command is not issued, then magic remains in a backwards-compatible mode.  Once the command is run, magic is no longer in the backwards-compatible mode, and scripts must be compatible with the new method of handling units, or some scripts are likely to break.  This is what happened with LibreLane.

I will issue a PR that comments out the line which will prevent random scripts breaking.  I apologize for letting that statement get into my pull request for magic support.

RTimothyEdwards:
This quick fix removes a line from the magic startup script that inadvertently breaks at least one LibreLane script and which was not meant to be in the last pull request for magic support.

Fixes #868 

- [ ] Tests pass

### DIFF
--- a/ihp-sg13g2/libs.tech/magic/ihp-sg13g2.magicrc
+++ b/ihp-sg13g2/libs.tech/magic/ihp-sg13g2.magicrc
@@ -54,8 +54,9 @@ source [file dirname [file normalize [info script]]]/ihp-sg13g2.tcl
 # load bind keys (optional)
 # source [file dirname [file normalize [info script]]]/ihp-sg13g2-BindKeys
 
-# set units to microns
-units microns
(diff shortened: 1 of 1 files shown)


KrzysztofHerman:
@RTimothyEdwards  thank you for this fix. The `main` branch is protected so it is not possible to push directly to it.
I will do in a separate PR from dev. I have tested the fix and it works fine.

PR #877, opened 2026-03-16
fix: correct typo in sg13g2_io.cdl and cell name in GDS

trabdelbasset:
- Fix subcircuit name from sg12g2_Gallery to sg13g2_Gallery in CDL
- Update top cell name in GDS from sg12g2 to sg13g2


KrzysztofHerman:
@trabdelbasset  thank you for detecting this notifying and fixing.  @b10346  please take a look it, looks reasonable.

PR #890, opened 2026-03-18
ihp-sg13g2: libs.ref: sg13g2_tech: Fix PITCH and OFFSET

dnltz:
LEF defines PITCH and OFFSET values with
  OFFSET {distance | xDistance yDistance}

Instead of using one distance for both, use the correct X and Y distances here. Also update the OFFSET pitch for X and Y values.

Changed the TopMetal1 pitch from 3.28 to  2.28 to match latest "make_track.tcl" files.


dnltz:
@b10346 can you please take a look? This patch updates the tech LEF file to match the make_track.tcl file values.

I updated TopMetal1 to `2.28` because that is what librelane is using. Let me know if that change was wrong!

https://github.com/IHP-GmbH/IHP-Open-PDK/blob/68c1265b1c4e5e1b4613bef34b9ce7bfe4b2c957/ihp-sg13g2/libs.tech/librelane/sg13g2_stdcell/tracks.info#L11

b10346:
I would say that changing pitch is technology LEF is not allowed. Original pitch values in technology LEF-file correspond defined routing grid for standard cell library and can not be changed. Using the grid with is differ than defined in technology LEF-file can cause difficulty for placement and for routing. It also leads to DRC violations and increasing design area.  So all related design and flow setup files should be created based on technology LEF values. Not the other way around.
Pitch 2.28 for TM1 leads to spacing DRC violation and can not be used. Minimal possible pitch is 3.28. It was changed in the latest library update.
It make sense to review all libre-lane setup files to be sure that it contains settings which correspond "golden" technology LEF from PDK.

dnltz:
> I would say that changing pitch is technology LEF is not allowed. Original pitch values in technology LEF-file correspond defined routing grid for standard cell library and can not be changed. Using the grid with is differ than defined in technology LEF-file can cause difficulty for placement and for routing. It also leads to DRC violations and increasing design area. So all related design and flow setup files should be created based on technology LEF values. Not the other way around. Pitch 2.28 for TM1 leads to spacing DRC violation and can not be used. Minimal possible pitch is 3.28. It was changed in the latest library update. It make sense to review all libre-lane setup files to be sure that it contains settings which correspond "golden" technology LEF from PDK.

Thanks for confirming 2.38 is a wrong value. I just opened a PR to fix this for LibreLane.

I'm a little bit confused because we use the same values in LibreLane/ORFS for generating  tracks. This update now adds the same values to the technology LEF-file. Are those values used for something else then routing tracks?

LEF file says if only one values is specified, it's used for X and Y. This is clearly wrong (see M1.b issue).

image

b10346:
Technology LEF defines metal routing direction and metal pitch.
The digital cells are designed for grid 0,48um - horizontal size, 0,42um - vertical size. Pin placement is optimized for the grid.
If the routing direction of any Metal is defined as horizontal, the pitch is vertical and can be described by one attribute definitely.
And vice versa.
This approach in tech LEF is used in IHP internal flow. I don not exactly understand why we should define the pitch for metal direction which is not allowed to use for routing. 

From other side, that is no contradiction in the fact that "make_trace" contains the string:

"make_tracks Metal1 -x_offset 0.0 -x_pitch 0.48 -y_offset 0.0 -y_pitch 0.42"

My understanding that this setting overrides LEF values and adds information: if M1 is going to rout somehow in vertical direction then use 0.48 um pitch for X distance between shapes.  The main point is that information about y_pitch 0.42 corresponds tech LEF data and does not overridden by "make_trace".  

I would prefer to keep tech LEF data as is (to minimize number of changes in golden), other words I would not include additional data in tech LEF which can cause user confusions (for example: if metal pitch definition contains X and Y, then it is possible to think that all routing directions are allowed). 
By the way, I have to investigate this subject more.

dnltz:
> Technology LEF defines metal routing direction and metal pitch. The digital cells are designed for grid 0,48um - horizontal size, 0,42um - vertical size. Pin placement is optimized for the grid. If the routing direction of any Metal is defined as horizontal, the pitch is vertical and can be described by one attribute definitely. And vice versa. This approach in tech LEF is used in IHP internal flow. I don not exactly understand why we should define the pitch for metal direction which is not allowed to use for routing.
> 
> From other side, that is no contradiction in the fact that "make_trace" contains the string:
> 
> "make_tracks Metal1 -x_offset 0.0 -x_pitch 0.48 -y_offset 0.0 -y_pitch 0.42"
> 
> My understanding that this setting overrides LEF values and adds information: if M1 is going to rout somehow in vertical direction then use 0.48 um pitch for X distance between shapes. The main point is that information about y_pitch 0.42 corresponds tech LEF data and does not overridden by "make_trace".
> 
> I would prefer to keep tech LEF data as is (to minimize number of changes in golden), other words I would not include additional data in tech LEF which can cause user confusions (for example: if metal pitch definition contains X and Y, then it is possible to think that all routing directions are allowed). By the way, I have to investigate this subject more.

@b10346 thank you for explaining it. I just wanted to understand why this can't be changed.

I'm totally fine with whatever is your decision here.

maliberty:
>  I don not exactly understand why we should define the pitch for metal direction which is not allowed to use for routing.

You can route wrong way and there should be a grid for such edges.  Nothing in the tech lef disallows wrong way segments.

b10346:
I would put this pull request to "on-hold". Need a time to make experiments.

maliberty:
ping

b10346:
It seems to be reasonable to have (X, Y) pitch format for the grid where horizontal and vertical pitches are not equal.
We will approve the PR after following corrections:
1) TM1 pitch must be 3.28. This is DRM requirement (minimal width and spacing section),
2) LEF standard format requires to have a space between value and semicolon:
 OFFSET	0.0 0.0;   => OFFSET	0.0 0.0 ;

Please fix it.

dnltz:
> It seems to be reasonable to have (X, Y) pitch format for the grid where horizontal and vertical pitches are not equal. We will approve the PR after following corrections:
> 
>     1. TM1 pitch must be 3.28. This is DRM requirement (minimal width and spacing section),
> 
>     2. LEF standard format requires to have a space between value and semicolon:
>        OFFSET	0.0 0.0;   => OFFSET	0.0 0.0 ;
> 
> 
> Please fix it.

@b10346 done

b10346 (approved):
looks correct

dnltz:
Thanks. I will send a PR for CMOS5L as well.

PR #908, opened 2026-03-26
Fix magic fill generation (2nd pass)

### LINKED ISSUE #880: Incorrect fill when using magic for fill generation
The fill generation rule deck for magic had been developed early on for the first test shuttle run, and was validated at the time.  Since then, however, work on a different process made it clear that the fill generation is much easier to manage if it uses a tech file with 1:1 mappings of the mask layers.  That is generally not done in magic because it makes it impossible to do connectivity searches and extraction;  however, since fill generation only needs to look at the geometry, it works best with the 1:1 mapping.  But that work was done after the test tapeout and was not validated.  In fact, it is rather badly broken.

The main thing broken is that there is are aliases `ALLM*` for all mask types belonging to a metal layer, but the aliases are used with the incorrect name `ALLMET*`, so all metal layers are getting ignored during fill generation.

A bit more minor error is that two layer names from the original deck before the translation to 1:1 mapping were used but are not defined and need to be changed to the equivalent mask layers.

I have made the appropriate fixes and will generate a PR for the update.

RTimothyEdwards:
This PR fixes a few remaining issues in the fill generation deck for magic (see issue #880).
Mainly, the distance rules for poly fill had been set to be the same as the distance rules for diffusion fill, when there are several key differences.  Those differences have been incorporated into this pull request, along with the implementation of a keep-out guard band around pwell-block layer edges, which had previously been flagged as "to do". 

- [ ] Tests pass

### DIFF
--- a/ihp-sg13g2/libs.tech/magic/ihp-sg13g2-GDS.tech
+++ b/ihp-sg13g2/libs.tech/magic/ihp-sg13g2-GDS.tech
@@ -827,12 +827,23 @@ style patternfill variants (),(tiled)
 	and-not	well_shrink
 
 #------------------------------------------------------------------------
+# A similar guard band is needed around PWELLBLK, at 1.5um distance
+# Rule AFil.i
+#------------------------------------------------------------------------
+
+ templayer	pwblk_shrink PWELLBLK
+	shrink  1500
+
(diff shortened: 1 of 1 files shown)


mole99:
@KrzysztofHerman I tested the changes successfully on my end.

PR #920, opened 2026-04-02
[LVS] Improve hierarchical (deep-mode) extraction of taps

martinjankoehler:
This is originated by @klayoutmatthias and did help us a lot (@simi1505) during our tapeout.

The problem was that
- Assume e.g. we have 2 guard rings in the schematic
- In the Layout, we overlap these two guard rings (east / west side)
- Then KLayout extracts a smaller area than given in the schematic
- This results in a LVS mismatch

During our discussions with @klayoutmatthias, he explained that there is an important concept called *cheats* (might sound phony but that is misleading) that should be employed by the LVS script to allow proper extraction (at least for `deep` mode).

This is tapeout phase verified by @simi1505.


simi1505:
@martinjankoehler This PR is included in PR https://github.com/IHP-GmbH/IHP-Open-PDK/pull/1032. Please close this PR. Thanks!

PR #930, opened 2026-04-13
KLayout PCell Libraries: properly register technology

### LINKED ISSUE #929: KLayout PCell registration: missing tech association
In the KLayout scripting API, there is `pya.Library.technology=`, which should be set so that the correct library can be shown.
This is especially important with the introduction of the `ihp-sg13cmos5l`:
https://www.klayout.de/doc-qt5/code/class_Library.html#method38
This ensures that the correct library is used for the layout.

martinjankoehler:
Fixes #929

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_native_pcell_lib/__init__.py
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_native_pcell_lib/__init__.py
@@ -23,9 +23,9 @@ from .via_pcell import ViaPCell
 class SG13G2_ViaLib(pya.Library):
     def __init__(self):
         self.description = "SG13G2 Native PCells"
+        self.technology = 'sg13g2'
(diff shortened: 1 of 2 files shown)


Mauricio-xx (approved):
Fix looks good. The order matters because `register_lib` uses the `(name, technologies)` pair to decide whether to replace an existing library; registering before setting technology means it registers with an empty set, breaking deduplication on reload.

The missing `self.technology` in `sg13g2_pycell_lib` is the more impactful change: without it, `SG13_dev` was visible for all technologies, which causes real problems when `ihp-sg13cmos5l` is also installed (both register under `SG13_dev`). See IHP-GmbH/ihp-sg13cmos5l#37.

@KrzysztofHerman lgtm

ThomasZecha:
pycell_error
@martinjankoehler: After some investigation I found out that the error is caused by your change:
git-diff

martinjankoehler:
@ThomasZecha I've created a PR to fix this: #934

PR #822, opened 2026-02-16
PyCell for Klayout

ThomasZecha:
-added initial support for pcell callbacks

PR #933, opened 2026-04-20
pycell_test: fix crash introduced by #930

Mauricio-xx:
Fixes the regression reported by @ThomasZecha in https://github.com/IHP-GmbH/IHP-Open-PDK/pull/930#issuecomment-4275611014.

PR #930 added `self.technology = 'sg13g2'` to `PyCellLib`, which restricts the `SG13_dev` library to layouts that have that technology set. The test was creating a bare `pya.Layout()` with no technology, so `create_cell()` was returning `None` for several cells and crashing on the first `DCellInstArray()`

Fix is one line: set `layout.technology_name = "sg13g2"` right after creating the layout.

Tested with:
```bash
cd ihp-sg13g2/libs.tech/klayout/sg13g2_tests
KLAYOUT_PATH=$(pwd)/.. klayout -zz -r pycell_test.py
```

Note: there are two unrelated non-fatal errors still showing up (from #822, `numeric` module and a Tcl callback issue). They don't affect the GDS output.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/sg13g2_tests/pycell_test.py
+++ b/ihp-sg13g2/libs.tech/klayout/sg13g2_tests/pycell_test.py
@@ -26,6 +26,7 @@
 # KLAYOUT_PATH=$(pwd)/.. klayout ihp-pycells.gds -e
 
 layout = pya.Layout()
+layout.technology_name = "sg13g2"
 pcellNmos = layout.create_cell("nmos", "SG13_dev", { "l": 0.350e-6, "w": 6e-6, "ng": 3 })
 pcellPmos = layout.create_cell("pmos", "SG13_dev", { "l": 0.350e-6, "w": 6e-6, "ng": 3 })
 pcellCmim = layout.create_cell("cmim", "SG13_dev", {})


ThomasZecha:
> Note: there are two unrelated non-fatal errors still showing up (from #822, `numeric` module and a Tcl callback issue). They don't affect the GDS output.

@Mauricio-xx: Please can you share details about that issues.

Mauricio-xx:
Hey @ThomasZecha, looked into the two non-fatal errors mentioned in this PR:

**Error 1:** `Invalid Json syntax: 'missing value to go with key'`

The error name is misleading, the JSON parses fine. The actual problem is in `parameters.tcl`; `setTechParameters` receives the Python dict as a string and iterates it with `dict for {key value}`. If any tech param value contains spaces, Tcl ends up with an odd number of tokens and throws that error. `dlo.py` catches it and re-raises it as "Invalid Json syntax", which makes it a bit hard to trace.

**Error 2:** `No module named 'numeric'`

The callbacks spawn a subprocess `python3 -c "import callback; ..."`. `callback.py` does `from numeric import *`, and `numeric.py` exists under `pycell4klayout-api/source/python/cni/`. The PYTHONPATH for that subprocess is built from `dispScriptFile` in `parameters.tcl`, but if that path doesn't include the `cni/` directory, the import fails.

Both are non-fatal; they only affect parameter coercion in the KLayout GUI, not `genLayout`. The test exits 0 and writes the GDS correctly. Just wanted to give you the details in case it's useful.

ThomasZecha:
@Mauricio-xx: Thanks for the info, I will fix it.

ThomasZecha:
> Error 1: Invalid Json syntax: 'missing value to go with key'

Fixed by https://github.com/IHP-GmbH/IHP-Open-PDK/pull/950

ThomasZecha:
> Error 2: No module named 'numeric'
> 
> The callbacks spawn a subprocess python3 -c "import callback; ...". callback.py does from numeric import *, and numeric.py exists under pycell4klayout-api/source/python/cni/. The PYTHONPATH for that subprocess is built from dispScriptFile in parameters.tcl, but if that path doesn't include the cni/ directory, the import fails.

@Mauricio-xx: The path to the cni/ directory is set via setCniPythonPath, see paramerts.tcl, so how could you produce this error?

Mauricio-xx:
@ThomasZecha You're right, my earlier description was incomplete. Error 2 is actually a downstream effect of Error 1.

setCniPythonPath runs at dlo.py:241, right after setTechParameters at line 240, inside the same try. When setTechParameters throws (Error 1), the except re-raises as "Invalid Json syntax" and line 241 never executes, so cniPythonPath stays at its initial "" (parameters.tcl:25). Subsequent PCells skip re-init, but their callbacks still fire (_callbackMap was populated before the throw), hit Stof, which calls iPDK_engToSci, and the subprocess gets PYTHONPATH=<callbacks_dir>: with cni/ missing, so from numeric import * fails.

Sorry for the misleading framing in the previous comment; I hand-waved on the cni/ part instead of tracing the actual cause.

PR #934, opened 2026-04-20
[KLayout PyCell Tests] Fix failing test case by assigning technology to Layout object

martinjankoehler:
Fixes issue: https://github.com/IHP-GmbH/IHP-Open-PDK/pull/930#issuecomment-4275611014

Now that we've registered the library with the proper technology association,
we also have to do this for the Layout created in the test case.


martinjankoehler:
I just saw this is a dupe of https://github.com/IHP-GmbH/IHP-Open-PDK/pull/933

PR #946, opened 2026-04-24
LVS deck switch for disabling A/P comparison of taps

KrzysztofHerman:
This PR disables by default the comparison of A/P for `ptap1/ntap1` devices. 
It can be switched on using `--tap_ap_compare` CLI switch or by checking an option in GUI. 

image

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/lvs/run_lvs.py
+++ b/ihp-sg13g2/libs.tech/klayout/tech/lvs/run_lvs.py
@@ -483,6 +483,7 @@ def generate_klayout_switches(
         "run_mode": run_mode,
         "no_net_names": "true" if args.no_net_names else "false",
         "spice_comments": "true" if args.spice_comments else "false",
+        "tap_ap_compare": "true" if args.tap_ap_compare else "false",
         "net_only": "true" if effective_net_only else "false",
         "top_lvl_pins": "true" if args.top_lvl_pins else "false",
         "no_simplify": "true" if args.no_simplify else "false",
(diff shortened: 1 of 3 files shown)


Mauricio-xx:
Hi @KrzysztofHerman, thanks for the PR. The CLI/GUI plumbing in `run_lvs.py` and the two macros looks fine, the option round-trips into `-rd tap_ap_compare=true|false` on the KLayout command line as expected.

What I think is missing is the deck-side consumer. The way other switches work in this flow, `run_lvs.py` forwards each `--switch` via `-rd` and `sg13g2.lvs` decodes them in the SWITCHES block (L101-179) into Ruby constants that the rule decks then read, like `COMBINE_DEVICES = bool_check?($combine_devices)` and so on. For `tap_ap_compare` this last step doesn't happen. `sg13g2.lvs` never decodes `$tap_ap_compare`, and `CustomTap` in `rule_decks/custom_devices.lvs:190-191` still hardcodes `enable_parameter('A', true)` and `enable_parameter('P', true)`. So the variable does arrive at KLayout, but no rule deck looks at it and the engine keeps comparing A/P on taps regardless of the flag.

I checked this empirically against the branch rebased on current `dev`. Took a copy of `ntap1.cdl` with the explicit A/P on R1/R2/R3 scaled 10x, ran it against the upstream `ntap1.gds`, and LVS fails identically with and without `--tap_ap_compare`. The official regression (`test-LVS-main`, `test-LVS-cells`, `test-LVS-SVS`) passes both on baseline and with the PR applied, but only because the current tap test cases all supply correct A/P, so they don't actually exercise the toggle.

I think there's basically two ways to handle this. If the deck consumer was meant to be part of this PR, the change would be to decode `TAP_AP_COMPARE = bool_check?($tap_ap_compare)` in the SWITCHES block and replace the two hardcoded `enable_parameter` calls in `CustomTap` (scoping should work, `CustomTap.new` runs after SWITCHES). On top of that a `tap_devices` regression case with intentional A/P mismatch plus a `.sw.yaml` exercising both modes would lock in the behaviour, and the `testing/README.md` note T1 would need a small update. If instead it's coming in a follow-up PR, it'd be cleaner to hide the user-facing surface for now, skip registering the GUI widget and mark `--tap_ap_compare` as `argparse.SUPPRESS`, so users don't see a flag that does nothing.

Small thing on the side: the commit subject ("Modifications in the lvs-deck...") could use a reword since no `.lvs` file is actually touched here.

simi1505:
It would also be nice to have an additional checkbox, "Disable `ntap` / `ptap` extraction", in the KLayout LVS. This would make life much easier for macro-level LVS as discussed in issue https://github.com/IHP-GmbH/IHP-Open-PDK/issues/955. For top-level LVS, including I/O cells, we still need a different solution. Thanks!

PR #949, opened 2026-04-24
Fix LU.b: report portions of N+Activ beyond 20 um max space

Mauricio-xx:
## Summary
- LU.b used `Region#not_interacting`, which filters whole polygons. A single N+Activ polygon with one end inside the 20 um halo of a ptap and the other end beyond 20 um was silently passed.
- Per DRM SG13G2_os_layout_rules.pdf, Section 7.2.2 (page 67, Figure 7.4), the rule applies to "any portion" of N+Activ to a pSD-PWell tie.
- Replace with `Region#not` (geometric subtraction of the 20 um halo), so LU.b returns the sub-polygon portions that lie outside the halo.
This mirrors LU.a's `ext_not` in `sg13g2_maximal.drc:2839`.


Mauricio-xx:
`drc_regression_cells` fails only on `sg13g2_pr/TSV`. Unrelated to this fix: `TSV` was removed from `sg13g2_pr.gds` by 737ab476b1b6e89618458b739af5df9b67e2d3b2 and re-introduced by 558209bb5debd46e58e3efcd315e7dc1c2462cc4, and it lacks the `LU.b` waiver that `nmos`/`pmos`/`dantenna`/etc. already have in `WAIVED_TESTS` (`testing/run_regression_cells.py`). Out of scope for this PR.

DiarmuidSLICE:
I would like to verify this but can get the correct version. Could you create a fork with this fix in it and I can gitclone that. 

Worked for me fine with the fork at: https://github.com/YacineSot/ihp-sg13cmos5l

It is very important LU rule is correctly modeled so I am very interested in verifying this. 

Thanks

DiarmuidSLICE:
verified on my side now. LU check for NMOS working correctly (previously on PMOS LU check was working). 

Good job!

when will this be merged into the main lib?

Mauricio-xx:
Hi @DiarmuidSLICE thanks for testing this. We still need to do some internal testing before merge. Will be in the dev upstream soon.

-Mauricio

PR #957, opened 2026-04-30
Getting Monte Carlo Sampling to work for xyce (LV)

deerbeck:
Suggestion to fix Monte-Carlo Simulation for the sg13g2_lv_mos to work with xyce for high performance parallel simulation.
Very open to suggestions, shown changes work as a workaround for me but I am no IHP-dev.
Changes needed for my research internship.


KrzysztofHerman:
@deerbeck thank you for this contribution. Would it be possible to sent us some testcases that actually make use of this extended libraries ?  Xyce simulator is not on our priority list because of limitations of the ADMS model compiler but still if we get a valid extensions to our models we are open to merge it.

PR #984, opened 2026-05-19
Fix bug in RF MOS model geometry scaling 

p-fath:
## Summary

This PR fixes multi-finger geometry scaling for LV/HV NMOS and PMOS PSP wrappers.

Primitive calls now use per-finger width and explicit multiplicity:

```spice
w='w/ng'
nf=1
mult='m*ng'
```

Source/drain geometry is also passed per finger.

## Motivation

I found this while simulating an RF oscillator with `rfmode=1`. The design worked with `rfmode=0`, but with `rfmode=1` it behaved completely incorrectly (way too much off) and did not oscillate.

Reproducer:

https://github.com/p-fath/iruwbtx7

Without this PR, the oscillator does not oscillate using `rfmode=1` at the cross-coupled pair.

## Scope

Updated LV/HV NMOS/PMOS wrappers and mismatch wrappers. The PR did not change any behaviour if  `rfmode=0`, but it fixed the faulty behaviour if  `rfmode=1`.

ESD MOS clamp devices are unchanged.

### DIFF
--- a/ihp-sg13g2/libs.tech/ngspice/models/sg13g2_moshv_mod.lib
+++ b/ihp-sg13g2/libs.tech/ngspice/models/sg13g2_moshv_mod.lib
@@ -92,7 +92,7 @@
               + l=l
               + nf=1 mult='m*ng'
               + as='(max(w/ng,wmin)*(2*z1+max(0,(ng-2)/2)*z2))/ng'
-              + ad='(max(w/ng,wmin)*z2/2*ng)/ng'
+              + ad='max(w/ng,wmin)*z2/2'
               + ps='(2*(max(w/ng,wmin)*(2+max(ng-2,0)/2)+2*z1+max(ng-2,0)/2*z2))/ng'
               + pd='(max(w/ng,wmin)+z2)'
               + dta=trise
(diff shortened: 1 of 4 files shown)


simi1505:
@KrzysztofHerman and @dwarning, could you please look into this? We have a student currently working on this UWB transmitter, so we would appreciate getting this resolved as quickly as possible. Thanks for your help!

dwarning:
The patch implies that making a multifinger device is the same as multiplying the entire one finger device and connect them parallel. And we have in mind that we calling the psp model with the instance parameter nf and m which are then handled separate.
In my opinion we have to check the true root cause on simple test cases e.g. overlap and junction capacitances, series resistances and transconductances of inner device and some other.
Switching to rf device use a different and much more complicated model with NQS effects and some other features.

dwarning:
@p-fath : I need a simple test case to reproduce your modifications. Your recommended reproducer complex is complex.

KrzysztofHerman:
@p-fath thank you for reporting this but it seems that the root cause is not in he instance parameters. If you review the table 2.5.1 from the [PSP model](https://www.cea.fr/cea-tech/leti/pspsupport/Documents/psp103p5_summary.pdf) you will see that the definitions of `W`, `nf` and `mul`,  which are consistent with our current implementation. Please provide a simple testcase for a 1 transistor when we can get the `DCOp`, `gm`, `ft` and all four `S-params`. This way we can take a closer look.

KrzysztofHerman:
@p-fath @simi1505 we have found some possible root causes. We will have to update the model cards. Thank you for this intent to fix but this PR won't be accepted.

PR #999, opened 2026-05-26
LibreLane related cleanup

jeras:
This is just a cleanup, there should be no functional change.

1. Removing code related to multiplexer technology mapping. LibreLane is not using it, and it is not necessary, Yosys can properly map multiplexers to standard cells without additional map files.
2. Cleaning up `config.tcl` files, unused/deprecated` variables.

I was not yet able to test this change, since I have not yet figured out a command line that would allow me to run LibreLane with this library on local storage (instead of using `ciel`).


jeras:
I was able to run a test with a local clone of the PDK (it took some effort to figure out the correct command line):
```
librelane --manual-pdk --pdk-root ~/VLSI/PDK/IHP-Open-PDK/ --pdk ihp-sg13g2 --scl sg13g2_stdcell test/designs/spm/config.json
```

jeras:
@mole99 as discussed, please review this commit

As a side not for future work, I noticed the ICG cell is excluded from synthesis.
I did not conduct any clock gating tests yet, but I do plan to do so soon.
While looking at the Yosys source code, I noticed Yosys should (might) be able to detect ICG automatically,
so there should be no need to specify ICGs with LibreLane variables.

This means changes to LibreLane code handling clock gating variables might be needed to allow Yosys to run `clockgate`](https://yosyshq.readthedocs.io/projects/yosys/en/latest/cmd/index_passes_techmap.html#cmd-clockgate) without explicitely specifying which gates to use. This changes might affect the default handling of clock gating. I will write an issue when I get through testing.

mole99 (approved):
As discussed, this looks good to me. Thanks for the cleanup!

@KrzysztofHerman this would be ready to merge as soon as the draft state is removed.

mole99:
If the `clockgate` pass can automatically use the cells that are detected as ICGs, that would be great!

KrzysztofHerman (approved):

PR #1000, opened 2026-05-28
Add ihp-sg13cmos5l

dnltz:
Add IHP's CMOS5L PDK as sub-module.

This is a requirement to add CMOS5L to CIEL. Additionally, it makes access to this PDK easier.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "ihp-sg13g2/libs.tech/palace"]
 	path = ihp-sg13g2/libs.tech/palace
 	url = https://github.com/VolkerMuehlhaus/gds2palace_ihp_sg13g2.git
+[submodule "ihp-sg13cmos5l"]
+	path = ihp-sg13cmos5l
+	url = git@github.com:IHP-GmbH/ihp-sg13cmos5l.git


dnltz:
This PR adds the most recent version of CMOS5L.

See https://github.com/fossi-foundation/ciel/issues/124 why this is required to add this PDK to CIEL.

dnltz:
@mole99 FYI

mole99 (approved):
Thank you, Daniel!

@KrzysztofHerman Could you please take a look?

KrzysztofHerman:
@dnltz @mole99 it is under evaluation.

KrzysztofHerman:
@dnltz @mole99 since we are introduce new pdk extensions like LDMOS, RRAM, Cryo models we are still discussing on the best method to expose this data to the public. For this reason I will convert this PR into draft mode.  If you have any suggestions on how we could handle the data in the future your opinion is more than welcome.

mole99:
@KrzysztofHerman That is understandable. However, please keep in mind that Ciel integration is blocking a number of tasks, for example:

- Ciel integration is needed so that we can add unit tests to LibreLane for ihp-sg13cmos5l.
- Once ihp-sg13cmos5l is in Ciel, we can update [ihp-sg13g2-librelane-template](https://github.com/IHP-GmbH/ihp-sg13g2-librelane-template) to work with both PDK variants.
- Users can download ihp-sg13cmos5l via Ciel in the same way as sky130[A|B], gf180mcu[A|B|C|D] and ihp-sg13[g2].

The sooner we have a common repository to reference both PDK variants, the better.

PDK extensions (basically directories in `libs.ref/`) are already supported in Ciel. You can build/enable the PDK with just the base set of libraries, with all libraries, or a selection of libraries.

dnltz:
@KrzysztofHerman I honestly think IHP should make CMOS5L an tech and merge everything into sg13g2. Technically SG13G2 is just CMOS7L + HBT. So, add cmos5l and cmos7l as technology to this repository and add optional features like HBT, RRAM, Cryo, ...

This would allow IHP to produce the entire variation. You have a base technology and can add multiple options.

dnltz:
This would be a little bit like RISC-V handles extensions. You have the base ISA and can add extensions.

mole99:
@dnltz Having a single PDK variant with different metal stack-ups is not as straightforward as you might think, and is probably not easily compatible with open_pdks.

The PDK variable defines the PDK variant that is used (ihp-sg13g2 or ihp-sg13cmos5l), and therefore the stack up etc.
It is used, for example, by LibreLane to load the correct LibreLane PDK configuration.
Having only a single PDK variant for IHP complicates things. For example, think of the I/O cells that need to use different metal layers. You would need some other way to select the stack-up.

- sky130 has two PDK variants: sky130A and sky130B, which differ in their stack-up due to the addition of ReRAM.
- gf180mcu currently has four different PDK variants in open_pdks: gf180mcu[A|B|C|D], due to all the different process options.

dnltz:
> @dnltz Having a single PDK variant with different metal stack-ups is not as straightforward as you might think, and is probably not easily compatible with open_pdks.
> 
> The PDK variable defines the PDK variant that is used (ihp-sg13g2 or ihp-sg13cmos5l), and therefore the stack up etc. It is used, for example, by LibreLane to load the correct LibreLane PDK configuration. Having only a single PDK variant for IHP complicates things. For example, think of the I/O cells that need to use different metal layers. You would need some other way to select the stack-up.
> 
>     * sky130 has two PDK variants: sky130A and sky130B, which differ in their stack-up due to the addition of ReRAM.
> 
>     * gf180mcu currently has four different PDK variants in open_pdks: gf180mcu[A|B|C|D], due to all the different process options.

okay good to know. So, adding ihp-sg13cmos5l next to the g2 PDK would be the right way. However, I'm not a way adding it as submodule.

Let's say we need to update Metal1 in the tech lef:

- Open PR in this repo to update LEF
- Open PF in SG13CMOS5l repo to update LEF
- Update submodule in this repo to pull in the latest update

this workflow just adds extra workload and makes changes slower. I'm also not a huge fan of adding symlinks in the CMOS5L repo that point to non-existing directions (without the g2 checked out in parallel).

Here's what I would propose @KrzysztofHerman :

* Add the entire source code of ihp-sg13cmos5l to this repo.
* Move all SRAM macros to `IHP-Open-PDK/common/libs.ref/`. Add hard symlinks to `ihp-sg13g2` and `ihp-sg13cmos5l`. 

Next, we can start to move more common content to `common/` and add symlinks. Everything that can't be shared stays physically in the PDK directory. This would allow to significantly reduce the amount of code.

mole99:
This sounds like a good solution to me.

So we would have:

```
IHP-Open-PDK/common/
IHP-Open-PDK/ihp-sg13g2/
IHP-Open-PDK/ihp-sg13cmos5l/
```

All of the common things like SRAM, standard cells etc. would be in `common/` and symlinked into the two other directories. Things that are not shared, such as I/Os, can be kept separate in each directory.

@KrzysztofHerman what do you think of Daniel's suggestion?

dnltz:
> This sounds like a good solution to me.
> 
> So we would have:
> 
> ```
> IHP-Open-PDK/common/
> IHP-Open-PDK/ihp-sg13g2/
> IHP-Open-PDK/ihp-sg13cmos5l/
> ```
> 
> All of the common things like SRAM, standard cells etc. would be in `common/` and symlinked into the two other directories. Things that are not shared, such as I/Os, can be kept separate in each directory.
> 
> @KrzysztofHerman what do you think of Daniel's suggestion?

 @mole99 exactly. We would keep the open_pdks structue in each PDK directory.
 
 @KrzysztofHerman I can also open a draft pull-request with some of those changes. Might be helpful for your internal discussion.

PR #1003, opened 2026-06-01
Update issue templates

sergeiandreyev:
Enabled 'Bug report' template using GitHub settings UI


KrzysztofHerman:
@sergeiandreyev this PR mixes a lot of stuff. Please make comments in the PR description on what it actually does.

sergeiandreyev:
This PR was initially created for the `main` branch (this is how Issue template configuration is working from the GitHub UI -> Settings -> General) and then moved to the `dev`.
why we see here a lot of files not related to the actual PR? because we have an unexpected discrepancy between `dev` and `main`:
image

maybe we could use this PR to understand what's going on and to make correct actions to align the data

sergeiandreyev:
@b10346, could you please take a look at the IO library update that is included in this PR? should we merge it to `dev`?

b10346:
> @b10346, could you please take a look at the IO library update that is included in this PR? should we merge it to `dev`?

I see just one PR regarding sg13g2_io. This is a fix of top cell. The top cell will be excluded from the library in the upcoming IO update. I makes no sense to merge into dev.

PR #1005, opened 2026-06-05
add Xyce netlists for Qucs-S symbols

KrzysztofHerman:
This commit adds explicit XyceNetlist entries across the updated Qucs-S symbol set

Worked:
- Added XyceNetlist definitions for bondpad, diodes, taps, substrate, MOS, RF MOS, BJTs, and MIM capacitor symbols.
- Preserved the existing Ngspice and CDL netlist definitions while mirroring Xyce-specific parameters such as mm_ok where the symbols already carried them.

Did not verify:
- No Qucs-S netlisting smoke test was run in this commit.
- Xyce simulation run was executed for rhigh, rppd, npn13G2, MiM, ptap1 devices.
- No broader library regression or integration validation was run as part of this commit.


abdur-rahman-dev (approved):
I have run  netlisting  smoke test. Everything looks ok. 

due to line break a space is created in netlist. but functionally it is not a problem. 

Image

PR #1009, opened 2026-06-09
PyCell for Klayout

ThomasZecha:
-fixed missing deletion of original ellipse in function dbConvertEllipseToPolygon
-updated submodule ref pycell4klayout-api


sergeiandreyev (approved):
Tested instantiation in `KLayout` (`circle` type) and DRC run

PR #1016, opened 2026-06-16
Update Sealring Generator

dnltz:
Swap width/length in the sealring_code.py script to match the standard: width defines the x-axis and length defines the y-axis. Additionally, update the sealring.py script and change height to length.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/tech/scripts/sealring.py
+++ b/ihp-sg13g2/libs.tech/klayout/tech/scripts/sealring.py
@@ -2,7 +2,7 @@
 Klayout's batch mode. For example:
 
 klayout -n sg13g2 -zz -r sealring.py \
-        -rd width=1300.0 -rd height=1300.0 -rd output=macros/sealring.gds.gz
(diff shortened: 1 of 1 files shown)


dnltz:
@KrzysztofHerman @sergeiandreyev would be great if someone can take a look at this PR.

KrzysztofHerman:
@dnltz same question as in CMOS5L, what standard do you mean ?

dnltz:
> @dnltz same question as in CMOS5L, what standard do you mean ?

okay, the convetion is that length defines the longer pair and width the smaller. I just remove the pycell changes and only fixed the sealring script. Maybe we should use `bottom` and `side` or something similar. Anyhow, still fixes the current issue with swapped values.

KrzysztofHerman:
@dnltz after internal discussion we have decided to not to accept this change because of backward compatibility

dnltz:
@KrzysztofHerman can I add length as alternative to height and both result in the same internal variable? The currently existing height argument would print a deprecated warning to inform everyone to use length at some point.

Issue #958, opened 2026-04-30
[LVS] MIM cap is not extracted correctly when SBD is sitting underneath


simi1505:
While working on our [Six-Port receiver](https://github.com/iic-jku/SG13G2_SPARX), we discovered that MIM caps. are not correctly extracted during LVS when there is a Schottky barrier diode (SBD) sitting underneath.

Attached, you can find a [lvs_test.zip](https://github.com/user-attachments/files/27247968/lvs_test.zip) file with two test cases with one 10um x 10um MIM cap in parallel to one SBD:

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/90796c44-30d3-4c89-b1f6-1432bf759e72" />

In `lvs_test1`, both devices are placed beside each other. Resulting in clean Klayout LVS and DRC.

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/f9105436-a2fa-4069-9e66-496454559039" />

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/e8d0f4e3-efb9-4f53-9a1a-236b64403296" />

In `lvs_test2`, both devices are placed underneath each other. Resulting in a clean Klayout DRC, but the LVS fails.

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/3dac7091-24e6-4c7e-ab0a-8eb15ba35477" />

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/f6f9a6a0-afbd-48db-92df-d9829b7608ab" />

From my understanding, this should not be an LVS issue, right? Therefore, I would see this as an existing LVS bug. Please correct me if I am wrong.
I already talked to @klayoutmatthias about this issue, and this is what he came up with:
For some reason, the LVS includes the `pwell.block` layer in the `mimcap_exclude` (see [cap_derivations.lvs](https://github.com/IHP-GmbH/IHP-Open-PDK/blob/main/ihp-sg13g2/libs.tech/klayout/tech/lvs/rule_decks/cap_derivations.lvs):29).
```
mimcap_exclude = pwell_block.join(rfmimcap_exc)
```
Technologically, this does not make sense to us? Perhaps it's a copy-and-paste error from the Well Caps. The MIMCaps shouldn't actually have anything to do with the PWell.

If we simply replace it with:
```
mimcap_exclude = rfmimcap_exc
```
then the LVS is clean.

[...]

PR #1030, opened 2026-07-03
[LVS] fixed "SBD under CMIM" issue and added testcase

simi1505:
### Problem

A regular `cap_cmim` fails KLayout LVS when another device is placed directly underneath it, even though DRC is clean and Magic plus Netgen LVS pass. The reported case is a 10x10 um MIM cap stacked over a `schottky_nbl1` diode.

The cause is in `rule_decks/cap_derivations.lvs`. as explained in https://github.com/IHP-GmbH/IHP-Open-PDK/issues/958, the RFCMIM device is recognized by the `pwell_block` layer, and the regular CMIM path subtracts `pwell_block` to avoid claiming an RFCMIM:

```
mimcap_exclude = pwell_block.join(rfmimcap_exc)
...
rfmim_area = pwell_block.interacting(mim_drw)
```

Two things make this too broad:

1. `rfmim_area` uses `interacting`, so any `pwell_block` that merely touches a MIM turns that area into an RF region.
2. `cmim_top = mim_top.not(mimcap_exclude)` subtracts `pwell_block` at the region level, so a foreign `pwell_block` carves a hole in a regular MIM plate.

A Schottky diode brings its own `pwell_block` (it is part of the Schottky recognition layers). When the diode sits under the MIM, that `pwell_block` only partially underlaps the cap, but it is enough to break the CMIM extraction. The MIM is not an RF cap, the deck just cannot tell them apart.

### Fix

Recognize RFCMIM only when the MIM marker is fully enclosed by `pwell_block`, using containment instead of interaction, and exclude only genuine RF wells from the regular CMIM:

```
rfmim_area = pwell_block.covering(mim_drw)
mimcap_exclude = rfmim_area.join(rfmimcap_exc)
```

For a genuine RFCMIM (MIM inside its `pwell_block` ring) `covering` and the old `interacting` select the same well, so that path is unchanged. Only the case of a `pwell_block` that partially underlaps a MIM behaves differently, which is the bug. The downstream RFCMIM derivations (`rfmim_top`, `rfmim_btm`, `rfmim_dev`, `rfmim_sub`, `rfmeas_mk`) keep using `rfmim_area` and are unaffected.

Note on the workaround proposed in the issue: replacing the line with `mimcap_exclude = rfmimcap_exc` removes the only thing that separates CMIM from RFCMIM, so a real RFCMIM would then also match the CMIM path. It only appears correct on layouts that contain no RFCMIM. This PR keeps the CMIM and RFCMIM separation intact.

### Regression test

Adds a unit test `cmim_sbd` under `testcases/unit/cap_devices` (a `cap_cmim` in parallel with a `schottky_nbl1` diode, stacked), taken from the reporter test case in the issue.

- `testcases/unit/cap_devices/layout/cmim_sbd.gds`
- `testcases/unit/cap_devices/netlist/cmim_sbd.cdl`

The test fails on the old deck and passes on the fixed deck, so it guards against a regression.

### Follow up (not in this PR)

A more robust long-term option is a dedicated RFCMIM marker layer or label plus DRC rules that forbid foreign devices inside a real RFCMIM only, as proposed by @klayoutmatthias in https://github.com/IHP-GmbH/IHP-Open-PDK/issues/958. However, that is a direct PDK change and is left out of this deck fix. @KrzysztofHerman, maybe you can discuss this internally?


dnltz:
@simi1505 can you pleas rebase this pull-request?

KrzysztofHerman (approved):

PR #1031, opened 2026-07-03
relocate stdcell symbols/schematics into libs.ref and add xschem view selection

abdur-rahman-dev:
## Summary
Relocates the sg13g2_stdcell symbols and schematics out of `libs.tech` into
`libs.ref`, and adds an Xschem schematic-hierarchy selection so users can switch
between subcircuit, custom, and empty schematics.
## Details
- **Relocate into libs.ref**: move stdcell symbols and schematics into
  `libs.ref/sg13g2_stdcell/`, split per tool:
  - `sym/{qucs-s,xschem}/` — symbol definitions (`.sym`, Qucs-S `.xml`)
  - `sch/{qucs-s,xschem}/` — schematics (`.sch`)
- **Xschem hierarchy selection**: add a `stdcell_select_view_dialog` and a
  `hierarchy_config` procedure in `xschemrc` to resolve the schematic a
  symbol descends into. Three modes are selectable via
  `IHP -> Select stdcell hierarchy`:
  - Subcircuit (default) — descends into `<cell>.sch`
  - Custom — uses `<cell>_custom.sch` if present, else falls back
  - Empty — uses `<cell>_empty.sch` if present, else falls back
- **Qucs-S**: repoint the `symbols_stdcell` symlink at the new `libs.ref`
  location.
- **Test gallery**: move the stdcell example/test schematics to
  `sg13g2_test_stdcells/`.
- **Docs**: add READMEs for `sym/` and `sch/` documenting the layout and the
  Xschem hierarchy selection. Paths use `$PDK_ROOT/$PDK`.
- Corrected formatting of xschemrc and xschem-menu file.
## Motivation
Keeping stdcell symbols and schematics in `libs.ref` aligns the layout with
the rest of the PDK reference data and makes the files discoverable alongside
the other `sg13g2_stdcell` views (cdl, gds, lef, lib, spice, verilog). The
view-selection feature lets users run flat simulations without descending into
the full cell schematic.


d-m-bailey:
Couple comments and questions.

I think the original intention was to put tools specific files in libs.tech, and non-tool specific files in libs.ref.
It seems like your proposal puts data specific files in libs.ref, and technology specific files in libs.tech.

> The view-selection feature lets users run flat simulations without descending into the full cell schematic.

Not sure what you mean by flat simulations. Do you mean create netlists with black boxes?
I think the functionality you're proposing is available by changing the `XSCHEM_LIBRARY_PATH` in the xschemrc file. Does your proposal providing a dialog to do this?

Whether or not the subcircuits are netlisted or not is determined by the `type` property on the symbol. `type=subcircuit` are netlisted while `type=primitive` are not. I wonder if you could assign a variable and use that to switch during netlisting. For example, `type=@stdcell_type` and then `stdcell_type=subcircuit` or `stdcell_type=primitive`

abdur-rahman-dev:
@d-m-bailey Thankyou for your feedback. 
 
Here by flat simulation, I mean the generation of netlist as `primitive`. The initial idea was to switch the `type`. but unfortunately the variable in `type `does not expand like `format`. Your comment have me to rethink that and I realize I  can always pass tcl proc/variable there. 

The  `schematic` param for symbol that can point different schematics. The dialog is to choose which .sch to point to.

sergeiandreyev:
Hi @d-m-bailey, some additional comment on this:
the idea was to use `libs.tech` as **tool data** directory indeed, but `libs.ref` main idea, instead of 'non-tool specific data', was to use it for libraries/cellsets location (basically Digital Design Kit (DDK)) which is aligned to workspace structure I was used in the past (industry). Our PDK is quite small so we have just one "cellset" per-library (STDCELL library - only `sg13g2_stdcell` cellset, IO library - only `sg13g2_io` cellset, etc.)
Inside the 'cellset' the structure is `<view>`-based; these EDA views are in standardized and (more or less) tool-agnostic formats. So basically for the `inv_1` stdcell we have Verilog, CDL, Liberty, etc. descriptions which are placed inside `libs.ref/<cellset>` folder. Schematic and symbol view of the `inv_1` is the required part of DDK as other views.
And this is where this exception with schematic and symbols arises - we do not have a standardized format for these views (EDIF is not an option unfortunately), so we had to create this `<view>/<tool>` additional level of hierarchy to place the files. This is done for these two schematic tools we're supporting in our PDK..

PR #1036, opened 2026-07-08
ci: waive tkinter pip failure in make env

Mauricio-xx:
tkinter was re-added to `requirements.txt` in 288cd9db, but it is not a PyPI package (it ships as the system package `python3-tk`). With pip's resolver the install is atomic, so pip aborts before installing anything and every CI job that runs `make env` (linting, DRC, LVS) fails.

This is a temporary CI waiver, it does not remove tkinter. The `env` target still runs the normal `pip install -r requirements.txt`, and only if that fails it emits a `::warning::` and retries with the `tkinter` line filtered out, so the remaining dependencies install and CI proceeds. A genuine dependency failure (for example an unavailable klayout pin) still aborts, because the retry keeps every other package.

Verified locally: `make env` and `make lint_python` both exit 0, the warning is shown, and all deps (klayout 0.30.5, flake8, pandas, ...) install.

Proper fix for later: install `python3-tk` via the system package manager in CI and either drop tkinter or replace it with a never-resolving PEP 508 marker.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,17 @@ REQUIREMENTS_FILE := requirements.txt
 $(TOP_DIR)/actions_venv:
 	@python3 -m venv $(TOP_DIR)/actions_venv
 
-# Install requirements	
+# Install requirements
+# tkinter is listed in requirements.txt but is not a PyPI package (it ships as the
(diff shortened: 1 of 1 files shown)


sergeiandreyev (approved):
Tested by running `make env` locally

PR #1039, opened 2026-07-15
Via refactoring to also support cmos5l with the same code

martinjankoehler:
With this PR I've refactored the code in `sg2`, so that the same code can be used for both layer stack variants.
- `__init.py__`: added `SG13_Tech.TECH_NAME`, which then is used at all places where a `lyt` KLayout tech name is required
- `sg13_native_pcell_lib`: moved into the `sg13_pycell_lib` directory, so that above `TECH_NAME` becomes easily accessible from both layer stack variants
- `sg13_tech_info.py‎`: encapsulates the tech info regarding layers and vias and their design rules, this is used by the PyCell and Native PyCell, and now supports both `sg2` and `cmos5l`

I'll prepare a second PR for the `ihp-sg13cmos5l` repo, that more or less just symlinks to the `sg2` PyCell.

The changes from https://github.com/IHP-GmbH/ihp-sg13cmos5l/pull/47 should then be implemented here in this repo, so both layer stack variants benefit from the same code (and also a single place of maintenance).


ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
```diff
@@ -22,6 +22,7 @@
 # Creates the SG13_dev technology
 from .sg13_tech import *
 from .sg13_tech_info import *
+from .native_pcells.via_pcell import ViaPCell
```
dnltz, line 25: **Claude**: suggestion: the new `native_pcells/` directory has no `__init__.py`, unlike its sibling `ihp/` package — so this import relies on implicit namespace-package behavior rather than the repo's convention of explicit packages. Recommend adding an empty `native_pcells/__init__.py` to keep the import robust to however KLayout loads the lib.

ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/sg13_tech_info.py
```diff
@@ -174,117 +175,141 @@ class TechInfo:
...
-                LayerInfo(29, 0, 'Via2'),
-                LayerInfo(30, 0, 'Metal3'),
-                LayerInfo(49, 0, 'Via3'),
-                LayerInfo(50, 0, 'Metal4'),
+    def instance_for_tech(cls, tech: Tech) -> TechInfo:
+        #
+        # NOTE: depending on the PDK, this SG13_Tech could be sg13g2 or sg13cmos5l
+        #
+        is_g2 = tech.TECH_NAME == 'sg13g2'
+        is_cmos5L = tech.TECH_NAME == 'sg13cmos5l'
```
dnltz: **Claude**: suggestion: if `tech.TECH_NAME` is neither `sg13g2` nor `sg13cmos5l` (a typo or a future variant), this silently builds a stack with no top-metal vias instead of failing. Consider an `else: raise` so an unexpected tech name errors loudly rather than producing a subtly-wrong layout.

ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/sg13_tech_info.py
```diff
@@ -365,3 +397,22 @@ def get_vias(self, bottom: LayerInfo, top: LayerInfo) -> List[Via]:
...
 
+
+class TechInfoFactory:
+    @classmethod
+    def tech_info_for_tech(cls, tech: Tech) -> TechInfo:
+        if not hasattr(cls, '_tech_info_by_tech_name'):
+            cls._tech_info_by_tech_name = dict()
+            
+        tp = tech.getTechParams()
+        tech_name = tp['techName']
```
dnltz: **Claude**: nit: this caches by `tp['techName']` (from the JSON), while `instance_for_tech` picks the variant via `tech.TECH_NAME` (the class attribute). It's fine as long as they agree, but they're two sources of truth for the same identity — keying the cache on the same value the variant logic uses would remove the chance of them drifting apart.

ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/native_pcells/sg13_klayout_tech_info.py
```diff
@@ -23,9 +23,12 @@
 
 import pya
 
-from sg13g2_pycell_lib.sg13_tech_info import (
-    TechInfo, 
-    ViaInfo, 
+from ..sg13_tech import SG13_Tech
```
dnltz, line 30: Can we use absolute imports instead of relative to this file?

dnltz (commented):
**Claude**: Nice refactor — collapsing both layer-stack variants onto one code path reads cleanly, and I verified the sg13g2 path (what this repo actually runs) is behavior-preserving: the layer/via lists, layer_choices (which excludes via/cut layers via non_via_layers), and the new conditional vt2 params all match the previous hardcoded behavior. The conditionally-unset self.vt2_columns/vt2_rows are safe because they're only read for TopVia2 vias, which the cmos5l stack doesn't define. Left three small inline notes — none are blockers.

This refactor adds no test coverage, and it's now much more testable than before.

-----

This review was written partly or completely by Claude, an AI assistant, on behalf of the account submitting it. AI-generated feedback can be wrong or miss context — please read it critically and verify before acting on it.

ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
martinjankoehler: Done

ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/sg13_tech_info.py
martinjankoehler: Done

ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/sg13_tech_info.py
martinjankoehler: Done

ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/native_pcells/sg13_klayout_tech_info.py
martinjankoehler: No, because this file is sym-linked in `ihp-sg13cmos5l`, and the PyCell library directory has different names:
- `sg13g2`: directory is called `sg13g2_pycell_lib`
- `sg13cmos5l`: directory is called `sg13cmos5l_pycell_lib`

But I've added a comment explaining this.

martinjankoehler:
@dnltz @KrzysztofHerman @sergeiandreyev all review points were addressed, please re-review and merge.

KrzysztofHerman (approved):

KrzysztofHerman:
@martinjankoehler thank you for this contribution.

PR #1049, opened 2026-07-22
Add cap_mom interdigitated MoM capacitor to SG13G2

Mauricio-xx:
This adds `cap_mom`, an interdigitated metal-oxide-metal capacitor, as a full device for SG13G2 on the Metal1..Metal5 thin-metal stack.

It ships the four device deliverables plus their wiring:

- KLayout PCell (`sg13g2_pycell_lib`), parameterised by `w`, `l`, the metal stack (`mmin`/`mmax`) and the feed style (`double`, `same`, `none`), with PLUS/MINUS terminals and a Recog.mom (99/39) recognition marker. Registered in the pycell library, the device gallery, the pycell smoke test and the netlist-import templates.
- Verilog-A compact model compiled to OSDI with OpenVAF, plus a 2-terminal ngspice wrapper subckt wired into `.spiceinit` and the cap corner chain, and its compile-script entry. The model is the RF equivalent circuit (area and feed capacitance, series R and L, skin effect and a substrate shunt).
- xschem symbol and an AC testbench.
- LVS extraction and compare: the device is recognised by its Recog.mom marker, each metal's pins connect only to that metal so the single-side stacked feed stays on distinct nets, and it extracts as a 2-terminal device through the shared `DeviceCustomMIM` class. It is folded into the CAP device group with a `cap_devices` unit testcase and a syntax-reference entry.

Feed variants: cap_mom is one parameterised device, not two separate cells. The `feed` parameter selects the variant across all deliverables (layout, symbol, model and LVS). The two complete 2-terminal versions are `double` (opposite-side feed, both pins on the top metal) and `same` (single-side feed, two plates stacked on adjacent metals); `none` is a layout-only bare array. The two 2-terminal variants are not electrically identical: the single-side `same` adds a feed capacitance that `double` does not (and uses a different series resistance), so they have a different total capacitance computed from the same model. In LVS the single-side stacked plates are kept on distinct nets by per-metal pin connectivity, and the unit testcase exercises both `double` and `same`.

Verification done on this branch: the PCell generates the correct M1..M5 stack (and the reduced stacks) with zero geometric DRC errors; the LVS device regression passes with the new cap_mom testcase (both feed variants), each extracts as a 2-terminal device on two distinct nets, and a negative control fails; and the model simulates in ngspice through the corner chain, matching the PCell capacitance label.

One known limitation: the coefficients are the nominal g2 values by metal-layer count. Process-corner and mismatch variation are not yet characterised, so cap_mom returns the nominal value in all corners and in Monte-Carlo. This is a documented caveat to be modelled from silicon (matched-pair or array test structures on a future run), not a claim that the device is corner-stable.


KrzysztofHerman:
@Mauricio-xx 

Things to improve:

- stick to the name `cap_cmom` everywhere
- symbol for xschem: 
1. change the geometry to identical as cap_cmim
2. implement mm_ok=1
3. align text layers to be compact but not overlapping with the geometry
4. implement an estimate of a value based on parameters
5. add the symbol to the gallery.sch 
6. add sg13g2_tests/ac_cap_mom.sch to the sg13g2_tests/IHP_testcases.sch in AC category
7. add m parameter
- symbol for qucs-s in XML format
- add cap_cmom to the gnucap libraries, add a testcase

KrzysztofHerman:
@Mauricio-xx 

Symbols:

1. the equations in xschem and qucs-s do not work
2. in qucs-s the qualifier "double", "single" could be a drop down list
3. Gnucap empty "cap_cmom_paramset.va", no actual module form mom
4. Add verbose log here in the discussion to show the implemented changes

Mauricio-xx:
Thanks for the detailed list, all four points are addressed in the latest push.

**Naming.** cap_cmom is now used throughout, including a few LVS deck section banners and runtime log lines that were still carrying the old uppercase name. The custom extractor file and class keep the short "mom" abbreviation on purpose, mirroring `custom_mim_extractor` / `MIMCAPExtractor`; happy to spell those out too if you prefer.

**xschem symbol.** Geometry now mirrors `cap_cmim`, `mm_ok=1` and the `m` multiplier are both in the template and format string, and the labels are compact and clear of the glyph. The value estimate is derived from the parameters and shows the exact model value for each feed topology; because the expression is piecewise it lives in a small Tcl proc (`cap_cmom_est`, in a new `xschem-devinfo` file sourced from `xschemrc`, the same way `cap_cmim` uses `mim_drc`) so it renders reliably instead of as raw Tcl. The symbol is in `gallery.sch` under its own "MoM capacitor" category, and `ac_cap_cmom` is wired into `IHP_testcases.sch` in the AC group. Pins are named `c0`/`c1` to match the other capacitors.

**qucs-s.** An XML symbol (`symbols/cap_cmom.xml`) mirroring `cap_cmim.xml` is included.

**gnucap.** `cap_cmom_paramset` builds cleanly and `tb_cap_cmom_typ` runs deterministically, reporting 20.34 fF for the 5x5 double-feed instance, matching the layout anchor. Its golden reference is committed, so `make check-capacitor` reports PASS alongside the other capacitor Verilog-A tests.

Mauricio-xx:
**1. Equations in xschem and qucs-s.**

xschem (commit 8fe7475c): the estimate previously came from a Tcl proc sourced from xschemrc, which did not render (the label showed raw Tcl), and that xschemrc edit is what produced the branch merge conflict. It now computes inline with the builtin `ev` proc, exactly like cap_cmim, so no auxiliary file and no xschemrc change. Two concrete gotchas were behind the non-render: xschem drops a `@ref` that is not followed by whitespace, and `ev` runs `expr $i` unbraced so an embedded quoted string compare throws. The inline form avoids both. It shows the model capacitance for the double-feed active area and matches the layout anchors: 5x5 (M1..M5) 20.34 fF, 7x7 48.8 fF, 10x70 843.9 fF. The feed mode is still printed on the symbol (`feed=...`), so the topology is visible; the on-symbol number is the area term, since the feed delta is a small additive correction that a symbol-level `expr` cannot select on a string value.

qucs-s (commit 5a4f6c5d): the C equation used `floor()`. The qucs-s equation engine is stock muParser 2.3.3, which has no `floor()/ceil()/round()` and silently returns 0 on a parse error, so the symbol was showing C=0. Replaced `floor(x)` with `rint(x-0.5)`, which muParser supports and which reproduces the model value over the drawn range. Verified against libmuparser directly: 5x5 20.34 fF, 7x7 48.8 fF, 10x70 843.9 fF, matching the xschem estimate.

**2. qucs-s feed as a dropdown.** Done in the same commit. The generic-component XML has no enumerated-parameter attribute, but the properties dialog renders a combobox when the allowed values are given as a bracketed list in the parameter Description. feed is now `[none,same,double]` with default double.

**3. gnucap cap_cmom_paramset.va empty / no mom module** (commit ff987d39): the module was in fact present (a `--dump` emits the full `module cap_cmom(PLUS, MINUS, SUB)`) and tb_cap_cmom_typ was passing, but the file was misnamed: it was called `_paramset.va` while holding no paramset, just an include, so it read as a stub. cap_cmom is a single behavioural compact module with no corner/mismatch spread, so unlike cmim it has no sp_capacitor primitive plus corner paramsets. Renamed it to `cap_cmom_module.va` with a header that states plainly it is a build shim including the shared Verilog-A (kept single-source so the gnucap plugin and the ngspice OSDI build compile identical model code); updated the Makefile target and the tb load path. It rebuilds clean and `make check-capacitor` still reports PASS on tb_cap_cmom_typ (20.34 fF).

**Merge conflict.** The branch no longer modifies xschemrc at all (restored to the branch point), so the xschemrc conflict is gone.

One extra follow-up (2228daa4): the .va model and PCell snap the tiling floor with a +1e-6 pitch epsilon so the layout, the C-label and the model agree at exact-pitch drawn sizes; the two on-symbol estimates now carry the same epsilon (anchors unchanged, exact-pitch 21x8.9 reads 228.8 fF in both xschem and qucs-s).

Commits, on top of the earlier round: 8fe7475c (xschem inline estimate), ff987d39 (gnucap module wrapper), 5a4f6c5d (qucs-s equation + feed dropdown), 2228daa4 (pitch-snap epsilon).

Mauricio-xx:
Follow-up on the qucs-s side, two changes since the last round.

**Polarized qucs-s symbol** (3e1ef48b). The shared `Capacitor.sym` drew both plates identically, so the qucs-s symbol did not distinguish the positive and negative terminals. `cap_cmom` now points at its own `cap_cmom.sym`: a straight plate marked `+` for PLUS and a curved plate for MINUS, matching the xschem convention, with PLUS drawn on the left. This is drawing only, PLUS stays first in PortSym order so the netlist node order is unchanged (PLUS, MINUS).

**qucs-s AC example** (13372dc0). Added `qucs-s/examples/ac_mom_cap.sch` as the qucs-s counterpart of the xschem testcase `sg13g2_tests/ac_cap_cmom`. Same high-pass RC and the same -3dB extraction, `C = 1/(2*pi*f_3dB*100k)`, run as an ngspice control block (`meas`/`print`) so it reports the same number rather than a different metric.

Both flows give the same result at the SPICE level. The device line is identical (same subckt, same node order, same parameters): the qucs-s netlister emits

```
XC1 in out cap_cmom w=10.0u l=70.0u mmin=1 mmax=5 feed=double subblock=0 m=1 mm_ok=1
```

which is the same instance xschem nets (`XC1 in out cap_cmom w=10.0e-6 l=70.0e-6 ...`, `in`=PLUS, `out`=MINUS). Running the qucs-s example through ngspice returns `C = 8.185e-13` (818.5 fF) at `f_3dB = 1.944 MHz`, the same value the xschem test reports (both `freq_at = 1.94443e6`). The only earlier discrepancy (818.5 vs 843.9 fF) was the extraction method, not the device: -3dB RC vs a direct low-frequency admittance read on the identical model.

simi1505:
@Mauricio-xx, @KrzysztofHerman, do you have a particular time plan for when this MOMCAP PR will be finished and merged into `dev`? Thanks! :-)

Mauricio-xx:
Two follow-ups this round, plus a correction to an earlier note.

The cap_cmom symbol is now a plain two-plate capacitor in both xschem (`sg13g2_pr/cap_cmom.sym`) and qucs-s (`qucs-s/symbols/cap_cmom.sym`), instead of the polarized one-curved-plate drawing. A MoM interdigitated capacitor is physically symmetric, so it should not carry the polarized-cap convention; that convention stays on cap_cmim. This is drawing only, the terminal order is unchanged (PLUS is still the first pin/port), so the netlist and LVS are unaffected.

Commits: 062f4c50 (symbol), 0caf78cd (qucs-s example).

Mauricio-xx:
> @Mauricio-xx, @KrzysztofHerman, do you have a particular time plan for when this MOMCAP PR will be finished and merged into `dev`? Thanks! :-)

Hi @simi1505 we are still doing some polishing, soon will be merge and port to cmos5l is on they way as well

KrzysztofHerman:
@simi1505 we are working heavily on this. The timeline is to have interdigitated and fringe MoM's in both PDK's as soon as possible (next weeks). It will be without statistics since we have recently dropped test structures on the recent [shuttle](https://github.com/KrzysztofHerman/IHP__TEST9203). I am evaluating the @RTimothyEdwards fringe capacitor using OpenEMS and Palace to double check the values of capacitances. Here go an example of 10x10 um one
image

The problem with EM solvers is that it takes really long time to converge due to the grid size.

KrzysztofHerman:
@Mauricio-xx These are comments from Lukas (gnucap libraries) 

- Add the `mm_ok = 0 from [0:0];` parameter to the cmom `paramset` for future mismatch variant implementation.
- To handle device multiplicity in a Verilog-AMS standard-compliant way, `$mfactor` should be implemented in `cap_cmom.va`. See `r3_[cmc.va:327](https://cmc.va:327/)-330` for reference.
- Add ngspice test tb_cap_cmom_typ.sp to Gnucap directory for cross-validation._

simi1505:
@arpadbuermen, FYI. IHP is adding their MOM capacitor soon. Are there also some changes required to simulate this MOM capacitor with VACASK as well? Thanks! :)

Mauricio-xx:
> @Mauricio-xx These are comments from Lukas (gnucap libraries)
> 
> * Add the `mm_ok = 0 from [0:0];` parameter to the cmom `paramset` for future mismatch variant implementation.
> * To handle device multiplicity in a Verilog-AMS standard-compliant way, `$mfactor` should be implemented in `cap_cmom.va`. See `r3_[cmc.va:327](https://cmc.va:327/)-330` for reference.
> * Add ngspice test tb_cap_cmom_typ.sp to Gnucap directory for cross-validation._

grand, thanks for letting me know; I will handle this now

Mauricio-xx:
@KrzysztofHerman Comments form Lukas were implemented in the last commit

KrzysztofHerman:
@Mauricio-xx I have merged the PR #1061 from Lukas. It fixes the ngspice test suite.

Issue #875, opened 2026-03-15
LVS fails for isolated NMOS using isolbox


engrvip123:
I am having trouble clearing LVS for layout that uses an isolated NMOS inside an` isolbox`. LVS comparison is failing with net mismatches.

Need help on how the bulk terminal of nmos should be connected in schematic and layout. Any advice or example would be greatly appreciated. 
Testcase scheamtic and gds attached. [nmos_iso.zip](https://github.com/user-attachments/files/26003087/nmos_iso.zip)

@FaragElsayed2

engrvip123:
<img width="1242" height="635" alt="Image" src="https://github.com/user-attachments/assets/ae091278-c4fc-4ca0-8e93-814bff92a440" />

<img width="1105" height="612" alt="Image" src="https://github.com/user-attachments/assets/d99490d5-bec7-481a-9608-b0f2bb1ac278" />

d-m-bailey:
@engrvip123 I think that in order to extract `ptap1` you'll need to add `sub!` on layer `TEXT 63/0`.

The `bn` text is also shorted to `VDD` on the nwell guardring. Is `bn` supposed to be the psubstrate?

There may also be rule problems.

@H-Ojiro any ideas?

d-m-bailey:
@H-Ojiro Thanks for looking into this.

@engrvip123 I think the schematic you sent is missing the `ptap1` device. We added one as shown in the original screen shot. We were able to get a clean LVS result by:
1. removing the `bn` text.
2. adding `sub!` text on the `TEXT 63/0` layer over the isolated ptap
3. modifying the extraction rules

@FaragElsayed2 @atorkmabrains 
The current klayout rules do not have a connection to the deep nwell layer, nbulay. 
Also the ntap connection to nwell is ignored when there's a diode recognition layer.
```
ntap = nactiv.and(nwell_drw).not(ntap1_mk).not(recog_diode).not(gatpoly)
ptap_all = pactiv.and(pwell).not(recog_diode).not(gatpoly)
```
By removing the `.not(recog_diode)` from the above lines and adding `nwell_drw` -> `nbulay_drw` connectivity, LVS passed for the given test case with the above layout and schematic modifications. Do you remember why `recog_diode` was excluded from the ntap/ptap calculations?

With the above rule changes, we get a clean regression test too.

@KrzysztofHerman FYI

engrvip123:
[...]

PR #1059, opened 2026-07-28
LVS: connect isolbox n-tub to its contacted guard ring

schwarzj99:
## Problem

The isolbox `I` terminal (the nBuLay tub) can never be named. It always extracts as an
anonymous net, so a design that biases its tub through a guard ring and declares it as a
top-level port cannot pass LVS.

## Root cause

The isolbox PCell paints `Recog/diode` (99/31) over the whole NWell ring, and both tap
derivations subtract that layer:

```ruby
# general_derivations.lvs
ntap         = nactiv.and(nwell_drw).not(ntap1_mk).not(recog_diode).not(gatpoly)
# tap_derivations.lvs
taps_exclude = ....join(recog_diode)
ntap1_exc    = pwell.join(psd_drw).join(taps_exclude)
```

So a *contacted* isolbox guard ring forms no tap at all. Its Activ/Cont/Metal1 lands on a
device-less net, which `simplify` then deletes along with any label placed on it. The `I`
terminal is left holding nBuLay-derived geometry only, and since there is no nbulay text
layer, nothing can ever name it.

## Fix

Derive the n+ ring inside the isolbox NWell ring and tie it to the `I` terminal — 5 added
lines, nothing removed:

```ruby
# diode_derivations.lvs
isolbox_tie = nact_nwell.and(recog_diode).interacting(isolbox_recog)

# diode_connections.lvs
connect(isolbox_i, isolbox_tie)
connect(isolbox_tie, cont_drw)
```

Scoping the derivation with `recog_diode` confines it to the isolbox ring, so an `ntap1`
placed inside the tub stays a separate net.

## Testcase

`sg13_dnwell_inv` in `unit/diode_devices` — a deep-nwell isolated inverter. The NMOS sits in
the isolated pwell with its bulk on the source; the PMOS well sits on the nBuLay and is
therefore at the tub potential. Per DRC rule `NBL.d` (min. PWell width between nBuLay and
NWell, *different net*, 2.2 µm) an NWell on nBuLay is necessarily the same net as the tub,
so the PMOS bulk cannot be at any other potential.

It runs on default switches — no yaml overrides — and **it gates this fix**. Without
`isolbox_tie` the ring Metal1 is deleted as dangling, taking the `VDD` label with it, so
both the isolbox `I` terminal and the `ntap1` tie lose their name and the comparison fails
outright rather than only on the port check:

| | strict | with `--ignore_top_ports_mismatch` |
|---|---|---|
| before this PR | FAIL | **FAIL** |
| with this PR | PASS | **PASS** |

The second column matters because `run_regression.py` appends `--ignore_top_ports_mismatch`
to every run, so a port-only failure would not be observable from the suite. This case fails
on the netlist itself, so it goes red in CI without the fix.

## Testing

- `run_regression.py` unit suite: **all testcases passed, 0 failures**, before and after.
- `testcases/sg13g2_cells` (80 variants, run manually since that directory has no harness):
  76 pass / 4 fail, **identical** before and after. Those 4 failures are pre-existing and may
  simply need switches my ad-hoc runner did not supply.

## Known limitation, deliberately out of scope

The deck has no `connect(nwell_drw, nbulay_drw)`, so it cannot see an NWell-on-nBuLay short.
Magic models this — `ihp-sg13g2.tech` puts `nwell`, `dnwell` and `isolbox` in a single
connectivity group — and `NBL.d` implies the same. I hit it on a real design: Magic reported
the tub merged with the PMOS well, KLayout did not. Adding that connect reproduces Magic's
extraction exactly, but it changes extraction for every isolated structure (`nwell_iso` feeds
bjt, esd and varicap), so it belongs in its own PR with its own regression analysis.

Related: `NBL.b`–`NBL.f` are documented in the DRC rule tables but not implemented in the
KLayout DRC deck — only `NBL.a` exists. `NBL.d` is precisely the rule that would flag an
independently-biased NWell inside a tub.


simi1505:
@d-m-bailey, here is the version without needing my `--disable_tap_extraction` PR. :)

d-m-bailey:
@simi1505 Does this solve the problem mentioned in #875 ?

Is the unconnected `sub` net intentional?

schwarzj99:
@d-m-bailey both examples from #875 pass with this PR.
nmos_iso:  had to remove the bn pin placed on the contact ring to avoid port mismatches between vdd and bn. Passes afterwards.
nmos_iso_ref:the isolbox instance shows an orange ValueError, cleared by opening the instance and clicking OK. This is cosmetic as it passes either way

On `sub`: do you mean it should be sub!, like in the schottky_nbl1.cdl testcase? Happy to change it, it passes either way.

d-m-bailey:
> On `sub`: do you mean it should be sub!, like in the schottky_nbl1.cdl testcase? Happy to change it, it passes either way.

AFAIK, spice does not differentiate between `sub` and `sub!`. That's a convention from Seimen's Calibre, I think.
In spice, you'll want a `.global sub` line somewhere if you're using `digisub` and `ptap1` to keep the substrate isolated.

If you're connecting the substrate to something, then you'll probably need to make the `sub` net an actual port.

schwarzj99:
I left out `.global sub` because it breaks the top port check: with it, the testcase only passes when 
`--ignore_top_ports_mismatch` is set. The intent here was a unit test that passes strictly, without that flag. Happy to amend it if you prefer, since the flag is enabled for the testcases globally anyway.

PR #1064, opened 2026-08-01
Fix seal ring corruption caused by a compound scaling factor in NLCB_w/NLCB_l

simi1505:
## Problem

`NLCB_w` and `NLCB_l` write the width and length parameters back with an appended `u` unit, but format the number with `Ftos`, which applies a scaling factor of its own:

```tcl
set tmpw [GridFix [expr $tmpw*1.0e6]]   ;# -> 1550.0 (um) iPDK_setParamValue w [Ftos $tmpw 3]u $cellId
```

`Ftos` scales values of 1000 and above, so `1550.0` becomes `"1.55k"`. Appending `u` yields `"1.55ku"` — a compound suffix. `cni.numeric.Numeric` reads only the first scaling character and returns 1550.0 instead of 0.00155, a factor `1e6` too large.

## Impact

`NLCB_w`/`NLCB_l` are used by the `sealring` PCell only. The MOS devices use `NLCB_mos_w`/`NLCB_mos_l`. The threshold is 1000 µm, so the library default of `400u` works and only seal rings of 1 mm and above are affected.

For those, the PCell computes with 1.55e9 µm, and the coordinates overflow the 32-bit database-unit range. The cell ends up with `INT_MIN` corners and degenerate polygons:

```
sealring dbbox: (-2147483.648,-2147483.648;78.2,53.2)   # expected (0,0;1600,1600)
```

The corruption is baked into the chip GDS and crashes the PDK filler downstream:

```
ERROR: In sg13g2_filler_ActGatP.lym: 'holes': Internal error:
       dbPolygonGenerators.cc:394 m_open.empty () was not true in Region::holes
```

This surfaced with commit https://github.com/IHP-GmbH/IHP-Open-PDK/commit/e94d99dd63598b6e35935b133605f1599a611bee by @ThomasZecha, which registered `sealring` for `NLCB_w`/`NLCB_l`. Before that commit, the parameters reached the PCell unmodified.

## Fix

Write the number without a scaling factor, since the caller supplies the unit. `Ftos`'s second argument means "round to 3 digits after the comma" and `Ftos` itself formats with `%.3f`, so the same format is used directly:

```tcl
iPDK_setParamValue l [format %.3f $tmpl]u $cellId
iPDK_setParamValue w [format %.3f $tmpw]u $cellId
```

This keeps the existing rounding behaviour and only drops the scaling factor.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/callbacks/mos_cb.tcl
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/callbacks/mos_cb.tcl
@@ -142,7 +142,7 @@ proc NLCB_l {} {
     
     set tmpl [Stof [iPDK_getParamValue l $cellId]]
     set tmpl [GridFix [expr $tmpl*1.0e6]]
-    iPDK_setParamValue l [format %.10g $tmpl]u $cellId
+    iPDK_setParamValue l [format %.3f $tmpl]u $cellId
 }
 
 #******************************************************************************************************
@@ -176,6 +176,6 @@ proc NLCB_w {} {
     
     set tmpw [Stof [iPDK_getParamValue w $cellId]]
     set tmpw [GridFix [expr $tmpw*1.0e6]]
-    iPDK_setParamValue w [format %.10g $tmpw]u $cellId
+    iPDK_setParamValue w [format %.3f $tmpw]u $cellId
 }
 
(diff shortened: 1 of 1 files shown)


simi1505:
FYI, @KrzysztofHerman, @ThomasZecha. Please look at this as soon as possible. Currently, the seal ring is broken.
Maybe also directly look at @dnltz PR https://github.com/IHP-GmbH/IHP-Open-PDK/pull/1016 where he addresses some other issues with the seal ring.
@dnltz, are there any other issues with the seal ring Python scripts? I guess you reviewed it with Claude already? ;-)
Please also simultaneously update these changes to CMOS5L.
Thanks a lot!

ThomasZecha:
@KrzysztofHerman: PR tested, it fixes the issue, please merge.

dnltz:
> FYI, @KrzysztofHerman, @ThomasZecha. Please look at this as soon as possible. Currently, the seal ring is broken. Maybe also directly look at @dnltz PR #1016 where he addresses some other issues with the seal ring. @dnltz, are there any other issues with the seal ring Python scripts? I guess you reviewed it with Claude already? ;-) Please also simultaneously update these changes to CMOS5L. Thanks a lot!

agree, would be nice if we can merge #1016 too!

simi1505:
@KrzysztofHerman, @Mauricio-xx, please merge this PR. The seal ring is still broken.

PR #1066, opened 2026-08-03
Fix SG13_dev PCell coerce failure on tech params containing spaces

### LINKED ISSUE #1065: SG13_dev PCell coerce fails on tech params containing spaces (surfaces as "No module named 'numeric'")
**Describe the bug**

Instantiating any `SG13_dev` PCell through the KLayout PCell library
(`ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib`) aborts during parameter
coercion on `dev`. The first coercion of a PCell triggers
`PCellWrapper._initializeCallbacks()`, which crashes with `missing value to go with key`
(re-raised as `Invalid Json syntax: ...`). As a downstream effect, callback-bearing
devices then fail with `ModuleNotFoundError: No module named 'numeric'`.

This affects every `SG13_dev` PCell (nmos, pmos, cmim, cmomi, res, diode, inductor,
bondpad, ...). The `sg13g2_tests/pycell_test.py` harness fails at its first
`create_cell(...)`.

**Suggested Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Critical

The basic Tcl-callback library (`SG13_dev`) is the PCell instantiation path shipped in
the PDK, and the failure hits the first coerce of any device. Not marked Critical because
the DRC/LVS sign-off decks run against committed static GDS and do not exercise the PCell
coerce path, so those sign-off flows are unaffected.

**Environment**
- Branch: `dev`
- KLayout: version pinned in `versions.txt`
- Reproduced headless (`klayout -zz`), no GUI required.

**To Reproduce**
```
git clone --branch dev --recurse-submodules https://github.com/IHP-GmbH/IHP-Open-PDK.git
cd IHP-Open-PDK/ihp-sg13g2/libs.tech/klayout/sg13g2_tests
KLAYOUT_PATH=$(pwd)/.. klayout -zz -r pycell_test.py
```
The run fails at the first `create_cell(...)`. The log shows
`Invalid Json syntax: 'missing value to go with key' ... in PCellDeclaration.coerce_parameters`,
and for callback devices `ModuleNotFoundError: No module named 'numeric'`.

**Expected behavior**

All `SG13_dev` PCells coerce and build without error.

**Root cause**

`dlo.py` forwards the tech-parameter dict into a Tcl proc as a Python dict repr:

`ihp-sg13g2/libs.tech/klayout/python/pycell4klayout-api/source/python/cni/dlo.py` (~line 240)
```python
PCellWrapper._tcl.eval(f"setTechParameters {parameters}")
```

`setTechParameters` in
`ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/callbacks/parameters.tcl`
walks that string with `dict for {key value} $techParams`. `dict for` requires an even
token count, but several shipped tech-param values contain spaces
(`sg13g2_pycell_lib/sg13g2_tech.json`):
```
"vendorName": "IHP Microelectronics GmbH",
"authorName": "(c) 2023 IHP Microelectronics GmbH",
```
(`getTechParams()` also surfaces `LBE_h = 'No value'`.) The whitespace split then yields
an odd token count, so Tcl raises `missing value to go with key`.

Cascade: the crash happens at `setTechParameters` (dlo.py ~line 240), one line before
`setCniPythonPath` (~line 241). Because `_initializeCallbacks()` aborts there,
`cniPythonPath` stays empty. Later, `iPDK_engToSci` execs `python3 -c "import callback"`
with a `PYTHONPATH` that no longer contains the `cni/` directory, so
`callbacks/callback.py`'s top-level `from numeric import *` cannot resolve `numeric`
(which lives at `cni/numeric.py`), producing `No module named 'numeric'`. The numeric
error is the symptom, not the root cause.

Isolated confirmation (no KLayout needed): feeding the exact Python dict-repr of the
shipped tech params to the buggy `dict for` under `tclsh` reproduces
`missing value to go with key` verbatim.

**Proposed fix**

Parse the Python dict repr with a quote-respecting regex so space-bearing values keep
their key/value pairing. A PR against `dev` follows this issue. Alternative (cleaner, but
touches the `pycell4klayout-api` submodule): have `dlo.py` serialize a proper Tcl dict
instead of passing `str(dict)`.

**Additional context**

Likely why this was not caught earlier: the DRC/LVS regressions and CI run against
committed static GDS and never exercise the PCell coerce path, and `pycell_test.py` is not
gating in CI. The Tcl-callback framework also appears to be mid-migration on `dev` (legacy
`callback.py` still coexists).

Two unrelated `sg13g2_pycell_lib` failures surfaced while reproducing. They are separate
from this issue and should not be conflated:
1. `cmim` with default params hits a Tcl `abs()` arity error inside its own device
   callback ("not enough arguments for math function abs").
2. `nmos` with float-typed dimensional params (e.g. `w=6e-6`) fails `float('6.0u')`; the
   basic devices expect string-with-unit params (`"6.0u"`).

Mauricio-xx:
Fixes #1065.

`setTechParameters` in `sg13g2_pycell_lib/callbacks/parameters.tcl` parsed the
tech-param dict repr with `dict for {key value}`, which splits on whitespace. Several
shipped tech params carry spaces (`vendorName`, `authorName`, `LBE_h = 'No value'`), so
the token count goes odd and Tcl raises `missing value to go with key`. That aborts
`_initializeCallbacks` one line before `setCniPythonPath`, leaving `cniPythonPath` empty,
which later surfaces on callback devices as `ModuleNotFoundError: No module named 'numeric'`.
The net effect is that every `SG13_dev` PCell fails at its first coerce.

This replaces the parser with a quote-respecting regex so space-bearing values keep their
key/value pairing. Behavior is unchanged for the values that already parsed; it only adds
correct handling for values with spaces.

Validation: a standalone `tclsh` run over the full shipped tech-param set (numbers, unit
strings, empty strings, and space-bearing values) parses every pair correctly, and
`sg13g2_tests/pycell_test.py` then instantiates the `SG13_dev` devices without the coerce
crash.

Note on CI: the DRC/LVS regressions run against committed static GDS and never exercise
the PCell coerce path, so this change is invisible to those checks (which is also why the
bug went unnoticed). Adding a gating PCell smoke test would close that gap and could be a
separate follow-up.

Scope: a single file, `parameters.tcl`. The two unrelated basic-lib failures noted in
#1065 (cmim `abs()` arity, nmos float-param typing) are out of scope here.

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/callbacks/parameters.tcl
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/callbacks/parameters.tcl
@@ -85,11 +85,24 @@ proc setTechParameters {techParams} {
 
     set techParameters [dict create]
 
-    dict for {key value} $techParams {
-        # remove decoration from key/value
-        regexp {'?([^',]+)'?:?} $key match undecoratedKey
-        regexp {'?([^',]+)'?,?} $value match undecoratedValue
-        dict append techParameters $undecoratedKey $undecoratedValue
+    # techParams arrives as a Python dict repr, e.g.
+    #   'key': 'value', 'key': 0.001, 'key': 'value with spaces', ...
+    # A plain "dict for" over this string breaks whenever a value contains a
+    # space (e.g. 'No value'), because the whitespace split then yields an odd
+    # number of tokens and Tcl raises "missing value to go with key". Parse
+    # with a regex that respects the quoting so that space-containing values
+    # keep their key/value pairing intact.
+    set pattern {'([^']+)'\s*:\s*(?:'([^']*)'|([^,]+))}
+    foreach {match key quotedValue bareValue} [regexp -all -inline $pattern $techParams] {
+        if {$match eq ""} {
+            continue
+        }
+        if {$bareValue ne ""} {
+            set value [string trim $bareValue]
+        } else {
+            set value $quotedValue
+        }
+        dict set techParameters $key $value
     }
 }
 


ThomasZecha:
@Mauricio-xx: There is already a solution for that kind of Python/Tcl parameter mapping problem calling `setCurrentCellParameters`. Please take over this approach for `setTechParameters` too.

Mauricio-xx:
@ThomasZecha thanks for the pointer, and you are right that `setCurrentCellParameters` is
where this is already solved. I am closing this PR anyway, because the premise behind it
does not hold.

I went back and re-checked the reproduction and it was contaminated by my local setup: my
KLayout home also has another PDK installed that registers a technology named `SG13_dev`,
and its parameter set contains a value (`LBE_h = 'No value'`) that is not part of sg13g2.
That value is what made the token count odd and raised `missing value to go with key`.
With a clean environment, `dev` builds the SG13_dev PCells fine, so the failure I reported
in #1065 is not something the PDK has.

What is actually left in `setTechParameters` is much smaller: over the 917 shipped tech
params the parse mispairs the two space-bearing values, so `vendorName` comes out as
`'IHP'`, `authorName` as `'(c)'`, and two spurious keys appear. No Tcl callback reads
those, so there is no functional impact, and it did not deserve the priority I gave it.

If you do want the `setCurrentCellParameters` encoding applied to tech params for
consistency, the place is `dlo.py` lines 236-240 in the pycell4klayout-api submodule,
mirroring what lines 306-317 already do, plus the matching decode in `parameters.tcl`. One
detail if you or someone else picks it up: the tech dict is not all strings (roughly 550
str, 362 float, 5 int), so it needs the same type check that `coerce_parameters` has,
otherwise `.replace()` fails on the first float. Happy to open that against the submodule
if you want it, but it is a robustness cleanup rather than a fix, and I did not want to
leave a PR open on a wrong premise.

Details in #1065.

ThomasZecha:
@Mauricio-xx: Thanks for the detailed clarification.

Issue #1070, opened 2026-08-03
KLayout PyCell Isolbox: diode_layer parameter is never honored


martinjankoehler:
The PCell `isolbox` specifies a parameter `diode_layer` (Yes / No, is realized using Lisp/SKILL style `t` / `nil` strings:
https://github.com/IHP-GmbH/IHP-Open-PDK/blob/72d79ed2453a8000a9541ca90e0f323591012b24/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/ihp/isolbox_code.py#L46

Then here this string is saved into variable `diode_layer`:
https://github.com/IHP-GmbH/IHP-Open-PDK/blob/72d79ed2453a8000a9541ca90e0f323591012b24/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/ihp/isolbox_code.py#L74

**PROBLEM:** And here, this always becomes true, because the string is neither `None` nor empty:
https://github.com/IHP-GmbH/IHP-Open-PDK/blob/72d79ed2453a8000a9541ca90e0f323591012b24/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/ihp/isolbox_code.py#L110

<img width="2804" height="2220" alt="Bildschirmfoto 2026-08-03 um 17 17 22" src="https://github.com/user-attachments/assets/6ad05de1-03ab-4bfa-b4f7-3e5f9e4aed68" />

KrzysztofHerman:
@martinjankoehler I have merged your PR. Thank you for contribution.

PR #1072, opened 2026-08-03
KLayout PyCell bondpad: Fixed evaluation of parameter fill

martinjankoehler:
Analog to #1071 / #1070

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/ihp/bondpad_code.py
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/ihp/bondpad_code.py
@@ -331,7 +331,7 @@ class bondpad(DloGen):
                         y = y+d
                 #  for metal
                 
-                if fill == 't' :
+                if fill:
                     dbDeleteObject(car(idmask))
             # if stack
         else :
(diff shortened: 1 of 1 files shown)


simi1505:
@martinjankoehler, can you please fix the following one-line issue and update this PR? Please also update this PR in our iic-jku fork. Thanks!

This fix revives a line that then crashes: with `stack='t'` and `fill='t'`, `genLayout` now raises `TypeError: 'Polygon' object is not subscriptable` at `dbDeleteObject(car(idmask))`. `idmask` is the bare `Polygon` returned by `dbCreatePolygon()`, and `car()` does `value[0]`, which cni's `Polygon` doesn't support. Before this PR, the always-false `fill == 't'` kept the line dead, so the crash was never visible. Net effect: the fill-exclusion mask still isn't deleted, just with a traceback now.

Suggested fix, matching what the circle branch of the same function already does:

```python
if fill:
    dbDeleteObject(idmask)
```

Verified headlessly (pip `klayout` driving `PCellWrapper`, iic-jku `dev` @ `d79a8146`): with `car()` the TypeError fires. With `dbDeleteObject(idmask)` the pad generates cleanly, and exactly one mask polygon is removed (9896 → 9895 shapes; `fill='nil'` output unchanged).

sergeiandreyev (approved):

PR #1090, opened 2026-08-10
fix(pycell): give each process its own preprocessing directory

### LINKED ISSUE #1087: KLayout PCell library registers zero PCells when two processes load an IHP technology concurrently
**Describe the bug**

`sg13g2_pycell_lib/__init__.py` preprocesses every PCell module into the system temp dir under a name derived only from the module name, imports it, and deletes it again:

```python
modulePreProcPath = os.path.join(tempfile.gettempdir(), f"{moduleName}_pre.py")   # L227
...
os.remove(modulePreProcPath)                                                      # L245
```

The name carries no process identity, `/tmp` is shared, and `ihp-sg13g2` and `ihp-sg13cmos5l` ship the *same* module names (`rfnmos_code`, `bondpad_code`, …). Two KLayout processes loading an IHP technology at the same time therefore write and delete the *same* file, and the loser dies with:

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/rfnmos_code_pre.py'
```

Because that exception escapes the `PyCellLib()` constructor called from `autorun.lym`, the whole library fails to register — **not one broken PCell, but zero PCells**. Two KLayout windows, or a batch DRC/LVS job next to an interactive session, are enough to trigger it.

**Suggested Priority**
- [ ] Low
- [ ] Medium
- [x] High
- [ ] Critical

Silent, total loss of the PCell library in a completely ordinary usage pattern, and the failure mode (an empty library) does not point at its cause. It is also load-dependent, so it looks intermittent.

**Environment**

- PDK: `IHP-Open-PDK` `dev` (verified against `c1d6f49e`)
- KLayout 0.30.10
- Python 3.12, Ubuntu 24.04 LTS
- Observed in the IIC-OSIC-TOOLS container image

**To Reproduce**

1. Start two KLayout processes with an IHP technology at the same time — either two with the same PDK, or one with `ihp-sg13g2` and one with `ihp-sg13cmos5l`, which collide because the module names are identical across the two PDKs.
2. Inspect the PCell library in both.
3. One registers its PCells; the other has none, and its log carries the `FileNotFoundError` above from `autorun.lym`.

Ten concurrent loads reproduced it reliably. A single interactive session plus a batch DRC or LVS run is the realistic everyday case.

**Expected behavior**

Concurrent processes do not interfere with each other, and each registers its complete PCell library.

**Additional context**

Giving the preprocessed file a per-process name is sufficient. Write, import and remove are strictly sequential within one process, so the PID alone makes the name unique even when a single process loads both PCell libraries:

```python
modulePreProcPath = os.path.join(tempfile.gettempdir(), f"{moduleName}_{os.getpid()}_pre.py")
```

Wrapping the removal in `try/except FileNotFoundError` is worth adding as a belt-and-braces measure, so that a missing temp file can never take the whole library registration down again. If you would rather not hand-roll the name, `tempfile.mkstemp()` is the more idiomatic route.

Note that `ihp-sg13cmos5l` ships its own copy of this `__init__.py` rather than a symlink into `ihp-sg13g2`, so both copies need the change.

Mauricio-xx:
PCell modules using `#ifdef` were preprocessed into `/tmp/<module>_pre.py`, a name derived only from the module name. Two KLayout processes loading an IHP technology at the same time write and delete the same file, and the loser dies with `FileNotFoundError` inside the `PyCellLib()` constructor. That constructor runs from `autorun.lym`, so the whole library fails to register: not one broken PCell, zero PCells.

The library now creates one `tempfile.mkdtemp()` directory per load and preprocesses into it. The path cannot be shared with any other process, PDK or user, and a single `shutil.rmtree` in a `finally` replaces the per-file `os.remove`, so nothing is left behind in `/tmp` even on the `sys.exit(1)` path.

Reproduced on `dev` c1d6f49e with 8 concurrent `klayout -zz` loads: 2 of 8 registered zero PCells. With the change, 12 concurrent loads pass three times in a row, and the layout from `pycell_test.py` is identical except for the GDS timestamps. The reproducer is added to `sg13g2_tests`, not wired into CI.

Draft because `ihp-sg13cmos5l` ships its own copy of this file and still needs the same change before #1087 is fully closed. `ihp-sg13g2` is no longer affected by a concurrent cmos5l process, but two cmos5l processes still collide.

Fixes #1087

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
@@ -36,6 +36,7 @@ import re
 import importlib
 import importlib.util
 import pathlib
+import shutil
 import tempfile
 import traceback
 
@@ -165,6 +166,19 @@ class PyCellLib(pya.Library):
         self.description = "IHP SG13G2 Pcells"
         self.technology = SG13_Tech.TECH_NAME
 
+        # Modules using '#ifdef' are preprocessed into this directory before
+        # they are imported. mkdtemp() hands every process its own directory,
+        # so concurrent KLayout sessions cannot overwrite or delete each
+        # other's files. That also covers the case of a second IHP PDK being
+        # loaded, which ships the very same module names.
+        preProcDir = tempfile.mkdtemp(prefix='sg13g2_pycell_')
+
+        try:
+            self.registerPCells(preProcDir)
+        finally:
+            shutil.rmtree(preProcDir, ignore_errors=True)
+
+    def registerPCells(self, preProcDir):
         tech = Tech.get('SG13_dev')
 
         processNames = []
@@ -224,7 +238,7 @@ class PyCellLib(pya.Library):
             modulePreProcPath = None
 
             if len(defines) > 0:
-                modulePreProcPath = os.path.join(tempfile.gettempdir(), f"{moduleName}_pre.py")
+                modulePreProcPath = os.path.join(preProcDir, f"{moduleName}_pre.py")
 
                 pyPreProcessor = preProcessor(modulePath, modulePreProcPath, definesSet, removeMeta=False, resume=True, run=False)
                 pyPreProcessor.parse()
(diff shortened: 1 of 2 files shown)


KrzysztofHerman (approved):
Tested with G2 and CMOS5L

PR #1096, opened 2026-08-11
Add *DLY assertion in verilog, add timescale and add minimal possible gene…

### LINKED ISSUE #462: SRAM macro models - A_DLY pin
The [SRAM documentation](https://github.com/IHP-GmbH/IHP-Open-PDK/blob/main/ihp-sg13g2/libs.ref/sg13g2_sram/doc/RM_IHPSG13_1P_64x64_c2_bm_bist.txt#L96) for the `A_DLY` pin simply states:
> Delay setting, adjustment of memory internal timings; recommended setting: Tie to 1

Its function does not seem to be further documented nor is it reflected in either the Verilog or Liberty models.  
Would it be possible to provide more information on its purpose and if it has externally observable differences (eg interface timings), also add it in the Liberty and/or Verilog models of the SRAMs?  

If this is not possible or would be too much effort I would recommend an assertion be placed in the SRAM Verilog model that catches it being tied to zero. 

We just caught this mistake in [Croc](https://github.com/pulp-platform/croc) and while the documentation clearly states it should be tied to 1, it is rather easy to miss at the moment.

### LINKED ISSUE #615: Wrong boundary layer/datatype in SRAM macro
The SRAM macro with 32-bit data width ([RM_IHPSG13_1P_512x32_c2_bm_bist](https://github.com/IHP-GmbH/IHP-Open-PDK/blob/dev/ihp-sg13g2/libs.ref/sg13g2_sram/gds/RM_IHPSG13_1P_512x32_c2_bm_bist.gds)) which was added ~3 months ago, has the boundary on the wrong layer/datatype.

It uses 235/4 (sky130 prBndry), whereas the other SRAM macros use 189/4 as the correct prBoundary.boundary.

### LINKED ISSUE #657: SRAM Verilog models missing `timescale` directive

b10346:
SRAM update from 11.08.2026

1) The `timescale 1ns/10ps directive was added to the Verilog models.
2) The Verilog models were modified to avoid tying the A_DLY and B_DLY ports to logic 0.
3) The notification was added in text documents.
4) SRAM 1P64x16 is added in the library


sergeiandreyev (approved):

b10346:
@dnltz Could you please check this update on your designs which use SRAM?

sergeiandreyev:
converting to draft, we're going to include SRAM checks into the PDK repo and recheck before merge

PR #1100, opened 2026-08-12
Add HBT support for Gnucap

deutzlukas-hub:
This PR adds support for the `sg13g2_hbt` model in Guncap.

Please focus the review on the HBT corner modules, device module wrappers, and paramsets in the `gnucap/models/` directory.

The implementation has been tested and cross-validated against the PDK's ngspice reference  implementation; see figures below. The implemented HBT test cases are based on the PKDs existing Xschem examples. 

The shown figures can be reproduced from the test reference outputs by running  `python python/plot_hbt.py` from the `gnucap` directory.
 
The reference files were generated using the following versions: 

Gnucap: snapshot 2026.07.29
modelgen: snapshot 2026.07.29
ngspice:  ngspice-46

Thanks for reviewing!

tb_dc_hbt_13g2_typ

tb_dc_mc_stat_hbt_13g2


felix-salfelder:
parameter integer Nx = 1 from [1:4];

in combination with

      .re = 7.13E+00 * pow(4 / Nx, 1) * corner_hbt.vbic_re;

looks suspicious. Sure?

The "_sim->clear_limit();" change (woraround, hack?) deserves attention and a separate commit with an explanation.

(This is not a complete review, just the eye catchers.)

Mauricio-xx:
Hi @deutzlukas-hub ;

I ran the HBT tests on my machine (gnucap 2026.08.07, ngspice 45+). AC, DC and the process Monte Carlo all pass. The three mismatch MC cases finish, but they do not match your 2026.07.29 goldens bit for bit. That looks like a gnucap snapshot difference, not a broken model.

A few things I would like fixed before merge:

1. The Makefile defines VBIC_SRC and HBT_PARAMSETS and then never uses them as prerequisites. Change a paramset and you keep the old .so.

2. In the npn13G2 mismatch set, .rbi lost * corner_hbt.vbic_rbx, and mm_vbic_tf is computed but never applied.

3. l and v only accept mm_ok in [0:0], while the wrappers still advertise the parameter.

4. npn13G2v_5t declares El as a localparam, so you cannot set it.

5. verilog-a/vbic_1p3/ has no LICENSE or NOTICE. r3_cmc does.

Thanks!

deutzlukas-hub:
@felix-salfelder and @Mauricio-xx, thanks for reviewing.

@felix-salfelder, I fixed the unintentional integer division. I imported the VBIC model into the `gnucap-models` repo, see PR#8. Let's discuss  "_sim->clear_limit();" there.

@Mauricio-xx, I addressed points 1, 2, 4, and 5.

For point 4, @KrzysztofHerman, do we have mm statistical parameters for the G2L and G2V HBT device variants?  

Thanks for the help!

KrzysztofHerman:
@deutzlukas-hub all the HBT's share the same statistics

deutzlukas-hub:
Added mismatch paramsets for the G2L and G2V using the same statistics as for the G2.

PR #1103, opened 2026-08-17
libs.ref: add sg13g2_stdcell_hv — thick-oxide (3.3 V) standard cell library

vcaeken:
## What

A thick-oxide (3.3 V) standard cell library, `sg13g2_stdcell_hv`: all 84 cells of `sg13g2_stdcell` rebuilt on `sg13_hv_nmos` / `sg13_hv_pmos` with the same topology, pin names and pin order, named `sg13g2_hv_*` so both libraries can appear in one netlist. Today the 3.3 V devices exist only inside the `sg13g2_io` pad ring; this fills the gap for synthesizable 3.3 V logic (always-on domains, sensor/actuator interfaces, level-shifted periphery).

The transform, applied uniformly: gate length 130 nm → 450 nm (`Gat.a3` minimum for the 3.3 V FET), PMOS width × 2.40 (re-centres the switching threshold at Vm/VDD = 0.5046, which is the property the thin-oxide sizing encodes), NMOS widths unchanged, source/drain parasitics recomputed. Cost measured on `inv_1`: input capacitance 2.20×, FO4 delay 2.66×.

## What's in the PR

* `libs.ref/sg13g2_stdcell_hv/{cdl,doc,gds,lef,lib,spice,verilog,sym,sch}` — mirroring the dev-branch `sg13g2_stdcell` layout (xschem views under libs.ref per 1c0757e0/c90378d8), including `sym/qucs-s` and `sch/qucs-s` and the `sym/`/`sch/` READMEs, so the Qucs-S view set matches the thin-oxide library's.
* `libs.tech/librelane/sg13g2_stdcell_hv/` — a complete LibreLane SCL (site `CoreSiteHV` 0.48 × 7.14, HV driving/tie/fill/decap/diode/CTS cells, tracks, latch/mux/tri-state techmaps, exclude lists, and the opt-in `sdfbbp_map.v`), plus an HV-conditional block in the PDK-level `libs.tech/librelane/config.tcl` and a per-SCL copy of `sg13g2_tech.lef`.
* `libs.tech/xschem/xschemrc` — two functional lines: the HV symbol dir on `XSCHEM_LIBRARY_PATH` next to the thin-oxide entry, and `::SG13G2_HV_SCH` for symbol→schematic resolution (`hierarchy_config` is hard-wired to the thin-oxide sch dir; happy to unify under one proc instead if you prefer).
* `libs.tech/klayout/tech/pymacros/sg13g2_stdcell_hv.lym` — optional autorun macro registering the HV GDS as a KLayout cell library (see questions).
* `libs.ref/.gitignore` — `!*.spice`, so the spice view is actually tracked (the root `.gitignore`'s `*.spice` was silently dropping it).

## View coverage

| view | coverage |
|---|---|
| SPICE / CDL / Verilog / xschem sym + sch | **84 cells** |
| GDS + LEF | **84 cells** — 66 by 1-D monotone retarget of the thin-oxide layout, 2 tie cells, 16 by per-cell generators; 17-track 7.14 µm rows on the `CoreSiteHV` 0.48 µm site |
| Liberty, **3 PVT corners** | **84 cells, 668 delay/transition tables each** — 74 with timing arcs (52 combinational, 9 flip-flops with clk→Q/setup/hold, 5 latches with en→Q and closing-edge constraints, 6 tri-states with data + `three_state_enable`/`three_state_disable`, 2 statetable ICGs with the full clock-gate model), `sighold` as `driver_type : bus_hold` with measured capacitance and leakage, 2 tie cells and 7 physical stubs with area and measured leakage. Corners: `typ_3p30V_25C` (`mos_tt`), `fast_3p60V_m40C` (`mos_ff`), `slow_3p00V_125C` (`mos_ss`), on identical slew/load axes so STA can interpolate across them |
| Qucs-S sym + sch | **84 cells** — 84 schematics, 84 component definitions, 45 shared gate shapes |

Every cell has both a layout and a Liberty entry at every corner, so nothing is stripped at install time. The two statetable ICGs (`lgcp_1`, `slgcp_1`) carry the full clock-gate model — statetable, internal state pin, `state_function` on GCLK, the `clock_gate_*` attributes, CLK→GCLK propagation, enable setup/hold, CLK `min_pulse_width` and per-state leakage. CharLib has no input form for a statetable, so they are measured directly, as are the tri-states (no Hi-Z concept) and the bus holder (no output pin). They remain in the SCL exclude lists because an ICG is instantiated deliberately rather than inferred — exactly as the thin-oxide SCL treats its own. `doc/ReleaseNotes.txt` lists every remaining gap with its reason.

## Verification

* **DRC** — the PDK's own decks, fixed invocation: **0 cell-rule violations** on the padded 2-row array and on a 4-row shared-rail array with mixed vertical neighbours and mirrored placements (what a placed block actually produces). The 7 density items (`M1.j`…`TM2.c`) are chip-level checks a metal-less stdcell harness always trips; the thin-oxide library behaves identically through the same harness (6). Each of the 16 generated cells additionally had to pass a per-cell gate before joining the library (`cell_verify.py`: LVS, modular **and** maximal KLayout decks in an abutted mirrored context, Magic, structure and pin checks) — 16/16.
* **Magic DRC** — **0 `NW.*`** under the *strict analog* N-well rules (`drc(full)`, euclidean), not just the relaxed `DigiBnd` digital ones. The remaining Magic flags are tool-interpretation differences on geometry both KLayout decks accept (`Cnt.c` contact overlap, wide-metal `M1.e` on PDN rails).
* **LVS** — per cell against the CDL: **84/84 match** (the four device-less `fill_*` cells via the runner's own `--ignore_top_ports_mismatch` relaxation, as for the thin-oxide fill cells).
* **Functional equivalence** against the thin-oxide library as golden reference: combinational 60 cells / 452 vectors PASS (12 high-Z states included), sequential 16 cells / 400 samples PASS (ngspice, PSP103 via OSDI).
* **View consistency** — schematics (netlisted through the installed symbols with this branch's xschemrc) and CDL both match the SPICE netlist per cell on pin lists and device multisets, 84/84 over 920 devices.
* **Liberty as data** (`verify_lib.py`, run per corner — **3/3 PASS**) — 84 cells, 668/668 tables populated at each corner, every `area` equal to the drawn boundary to 1 nm², 2 169/2 170 load-axis series monotone (one documented waiver at typ; fast and slow are clean), drive limits present on all 81 output pins, 14/14 sequential cells with delay + setup + hold, and class-aware checks for the 6 tri-state outputs, 2 clock gates and 1 bus-hold pin. OpenSTA reads the file without warnings, and an independent characterizer (lctime) agrees to a median 2.9 % on delays / 0.0 % on transitions over 3 132 STA-region table points.
* **Blocks** — the reference 8-bit counter and a tri-state shared-bus design harden RTL→GDS on the shipped library with no design-level LIB/corner overrides (flops mapped natively by `dfflibmap`, `$_TBUF_` onto `sg13g2_hv_ebufn_2`, DRC 0/0, LVS clean, non-vacuous max-cap/max-slew). Two `spi_slave` variants pass the full IHP signoff deck with zero geometry violations.

Both PDF reports in `doc/` document the derivation, the physical sign-off and the characterization methodology.

## Known limitations (details in `doc/ReleaseNotes.txt`)

* Not silicon-proven; everything is simulation against the PDK models.
* No `internal_power` on any cell — the characterization was scoped to timing and leakage, so STA power reports see leakage only.
* 11 of 282 signal pins sit off the vertical Metal1 track grid (12 others were widened onto tracks); route with `RT_MIN_LAYER: Metal1` or expect jogs — the LEF OBS covers all non-pin Metal1.
* `dlygate4sd2_1` / `o21ai_1` delay ratios shift (their thin-oxide gate lengths were below the thick-oxide minimum); decap cells store less per area.
* `verilog/sg13g2_udp.v` is a **symlink** to the thin-oxide library's copy, not a duplicate — the `ihp_*` UDPs are shared, so a second definition would collide whenever both libraries are read, and could drift. A flow still includes exactly one `*_udp.v`. (An earlier revision shipped no UDP file at all and called that deliberate; @b10346 pointed out on #1207 that the view then does not elaborate on its own, which was correct.)

## Questions for maintainers

1. **KLayout macro**: no stdcell GDS is registered as a KLayout library today — keep the autorun macro, move it, or drop it?
2. **xschem wiring**: keep the two-line `::SG13G2_HV_SCH` parallel, or generalize `hierarchy_config`/`symtype` (and the view-selection dialog) to cover both libraries?
3. **Mux maps** (@simi1505's structure review): `dev`'s thin-oxide SCL no longer ships `mux2_map.v`/`mux4_map.v`, while this SCL does and sets `SYNTH_MUX_MAP`/`SYNTH_MUX4_MAP`. Happy to drop them and let ABC map muxes, matching the newer thin-oxide style — say the word.
4. **`doc/` naming**: the thin-oxide library ships per-corner datasheets (`sg13g2_stdcell_<corner>.pdf`); this one ships a library report and a characterization-methods report. Per-corner datasheets (`sg13g2_stdcell_hv_<corner>.pdf`, three of them now) can be added alongside them if that pattern should be kept.

*(The earlier question about trimming the cell set is moot: all 84 cells now have every physical view.)*

## Revision history

| rev | head | what changed |
|---|---|---|
| 1 | — | initial submission: 84 cells in the netlist/schematic views, 68 drawn, 66 characterized |
| 2 | `792e79a` | LibreLane SCL, PDK config HV block, per-SCL tech LEF, Liberty drive limits (fixes a vacuous max-cap/max-slew check and an OpenROAD TritonCTS segfault — [OpenROAD#11165](https://github.com/The-OpenROAD-Project/OpenROAD/issues/11165)) |
| 3 | `865ae09` | N-wells rebuilt to the strict analog rules: Magic `NW.c1`/`NW.e1` false positives eliminated at the source, well layer only, no device change |
| 4 | `301b36d` | all 84 cells drawn — flops map natively through `dfflibmap`, `sdfbbp_map.v` demoted to an opt-in, sequential area on the reference counter −31 % |
| 5 | `179486c`, `7262b34` | `libs.ref/.gitignore` fix so the spice view is actually tracked; `sighold` characterized; both reports to rev. 5 |
| 6 | `bfff921` | the 6 tri-states characterized (direct measurement — CharLib has no Hi-Z concept) and enabled in the SCL |
| 7 | `abaa7be` | `doc/` sync: release notes and both reports brought to the shipped state |
| 8 | `41487ef` | the 2 statetable clock gates characterized — **all 84 cells now timed** |
| 9 | `357a5ad` | **fast** corner (3.60 V, −40 °C, `mos_ff`), all 84 cells |
| 10 | `ac0c261` | **slow** corner (3.00 V, 125 °C, `mos_ss`) — full PVT coverage; also corrects the antenna-diode leakage in all three corners (its measurement deck left a pin floating, so the operating point was under-determined) |
| 11 | `6b473d2` | Qucs-S schematic and symbol views, gated against the SPICE netlist |
| 12 | `62f4c27` | fixes the fast and slow Liberty sequential emission reported by @simi1505 — `fix_lib_seq.py` had never been run over the two new corners, so 21 pin functions per file carried `function : "Q = D"` instead of the state variable and neither corner would parse. Structural only: no timing, capacitance or leakage value changes. `verify_lib.py` now runs OpenSTA `read_liberty` as its first check |

Detail for each revision is in the comments below.

Provenance: generated and verified by the scripts at https://github.com/ChipDesign-BV/sg13g2_stdcell_hv (Apache-2.0, derived from `sg13g2_stdcell`, Copyright 2023 IHP PDK Authors). Every generator, verification suite and the two reports are there; the library regenerates end-to-end from the thin-oxide PDK views.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


simi1505:
FYI, @KrzysztofHerman and @Mauricio-xx.

vcaeken:
@simi1505 Thank you for the smoke test — extremely useful. Rev. 2 (792e79a) folds everything you had to hand-write into the PR itself; the "Rev. 2" section in the description has the details. The short version:

* `libs.tech/librelane/sg13g2_stdcell_hv/` is now shipped (site, cell maps, tracks, excludes), plus the HV-conditional corner block in the PDK-level `config.tcl` and a per-SCL copy of `sg13g2_tech.lef`.
* Your `sdfbbp_map.v` idea is in the SCL as the supported flop path — extended to all posedge FF types (enables via the scan mux, sync resets folded into D) and formally checked against the Yosys cell semantics with `equiv_induct`. Only `SYNTH_EXTRA_MAPPING_FILE` still has to be set at design level; LibreLane doesn't let a PDK set it.
* The liberty now carries per-pin `max_capacitance`/`max_transition` derived from the characterized table axes, so the TritonCTS crash is gone and the limit checks are no longer vacuous. Agreed the null deref is worth an OpenROAD issue — it should error, not crash.
* The 12 layout-less flops/latches are stripped from the installed liberty (they stay in the source repo until drawn), so liberty ⊆ LEF by construction.
* Magic NW.c1/NW.e1: root cause confirmed — Magic's tech has no DigiBnd relaxation and applies the analog 0.62 µm rules. Measured against the strict rule the cells are 25 nm short (65 of 68; mux4_1/slgcp_1 need a deeper well jog). A well-bottom-only layout fix — no device change, no re-characterization — is worked out and planned as a follow-up rev, which would make the library clean under Magic's strict interpretation too. Your remaining 24 errors are wide-metal M1.e on the PDN rails; KLayout implements the same rule and reports it clean.

Re-ran your counter setup end to end with no design-level LIB/corner overrides: 8 flops on `sg13g2_hv_sdfbbp_1`, zero unmapped cells, CTS clean, LVS clean, KLayout DRC clean.

vcaeken:
Follow-up, commit 865ae09: **the Magic NW.c1 false positives are gone at the source** — the N-wells now satisfy the strict analog HV rules outright, not just the DigiBnd digital ones.

Well-layer-only rebuild (`fix_well_nwc1.py` in the source repo): bottom edge 2.625 → 2.570 µm, deeper jogs over the two deep P-active bands (`mux4_1`, `slgcp_1`), lateral overhang 0.24 → 0.62 µm so the outermost cell of a row is enclosure-clean without a neighbour's merged well. Zero device changes, no re-characterization — only the GDS differs; liberty/LEF/CDL/SPICE/Verilog are untouched.

Verified: Magic reports 0 NW.* errors on the abutted library array; the KLayout modular and maximal decks (including NW.c1.dig) pass both DRC contexts; per-cell LVS 68/68; and the counter re-hardened end to end with `magic__drc_error__count` 1161 → **24** — the remainder is entirely wide-metal M1.e on the PDN rails, where KLayout's implementation of the same rule (`M1_e` 0.22) reports the layout clean. The block-edge NW.e1 flags are gone too.

Upstream reports filed from this work: The-OpenROAD-Project/OpenROAD#11165 (TritonCTS segfaults instead of erroring on a liberty without max_capacitance/max_transition, with a minimal repro), #1106 (modular KLayout deck doesn't implement NW.c1/NW.c1.dig — its "0 errors" was silent on this rule), #1107 (Magic tech lacks the DigiBnd relaxation — the tool gap that made this fix worthwhile).

vcaeken:
Milestone, commit 301b36d: **all 84 cells now have layouts** — the view-coverage gap that forced the scan-flop workaround is closed.

The 16 cells the 1-D retarget could not produce (the 8 dfrbp/sdfrbp flip-flops, 4 latches, `ebufn_2/8`, `lgcp_1`, `sighold`) are drawn by per-cell generators (committed in the source repo). The root cause of every retarget skip turned out to be the same: a p+ source finger butting into the VDD rail tap. The generators remove it before the retarget y-map, apply a per-cell insert split so every template band lands on the shipped positions, restore the finger at the `slgcp_1` butted-junction convention, and build the N-wells to the strict analog HV rules. No device W/L/connectivity deviates from the CDL anywhere; each cell passed LVS + both KLayout decks + Magic in abutted context before merging, and the merged library signs off clean (KLayout both arrays, Magic zero NW.*, per-cell LVS 84/84).

What changes for users:
* **`SYNTH_EXTRA_MAPPING_FILE` is no longer needed.** Flip-flops and latches map through `dfflibmap` / the latch techmap exactly like the thin-oxide library; the exclude lists now mirror the thin-oxide ones (`sdfbbp_1`, `dfrbp_2`). `sdfbbp_map.v` stays as a documented opt-in for all-scan-flop designs.
* The liberty ships all 75 cells with measured areas (nothing stripped — every liberty cell has a layout); LEF has 84 macros.
* Counter smoke test without the techmap: 8× `sg13g2_hv_dfrbpq_1`, sequential area 1699.89 → **1178.96 µm²** (−31% vs the sdfbbp workaround, no more tie-off overhead), **Magic DRC 0, KLayout DRC 0, LVS clean** — the full flow now completes with zero DRC errors from both tools.

Still not characterized (drawn, no timing arcs): the 6 tri-states, the 2 statetable clock gates, and `sighold` — unchanged from the original scope. Timing remains schematic-characterized (single typ corner); PEX-based re-characterization is possible future work.

simi1505:
Thanks @vcaeken! I tested it again:

Tested the library at head 301b36d by hardening an 8-bit counter macro (RTL2GDS, LibreLane Classic flow) from the iic-jku SG13G2 AMS chip template, inside IIC-OSIC-TOOLS (LibreLane v3.1.0.dev3, OpenROAD 26Q3-1297, Magic 8.3.589, KLayout 0.30.5, netgen). Result: the flow completes end-to-end with the library as shipped, except for one config-load issue (below).

**What passes** (single corner `typ_3p30V_25C`, 50 MHz, 200×100 µm die):

- Magic DRC: 0 errors — the strict N-well redraw in 865ae09 resolved the ~1.2k NW.c1/NW.e1 errors we saw with the draft revision; Magic and the KLayout decks now agree
- KLayout DRC: 0 errors
- Netgen LVS: circuits match uniquely (285 devices, 67 nets)
- STA clean (setup/hold/slew/cap), antenna clean
- Flops map natively via `dfflibmap` to `sg13g2_hv_dfrbpq_1` — no custom techmap needed anymore
- The TritonCTS segfault we hit with the draft liberty is gone with the per-pin `max_capacitance`/`max_transition` added in 792e79a
- For reference: counter cell area is 2307 µm² HV vs 787 µm² for the same RTL on `sg13g2_stdcell` (~2.9×)

**One remaining issue: no `spice/` views.** The SCL registration inherits `CELL_SPICE_MODELS = libs.ref/sg13g2_stdcell_hv/spice/sg13g2_stdcell_hv.spice` from `libs.tech/librelane/config.tcl`, and LibreLane validates that path eagerly, so any HV run dies at config load:

```
Path provided for variable 'CELL_SPICE_MODELS[0]' is invalid:
'.../libs.ref/sg13g2_stdcell_hv/spice/sg13g2_stdcell_hv.spice' does not exist
```

Workaround we verified: copying the CDL verbatim to `spice/sg13g2_stdcell_hv.spice` works — netgen parses the M-card CDL fine and LVS matches. So either shipping the CDL content as a spice view (like `sg13g2_stdcell` does) or pointing `CELL_SPICE_MODELS` at the CDL in the HV branch of `config.tcl` would make the library work out of the box.

Thanks for the quick turnaround on the earlier gaps. The library is very usable now.

vcaeken:
Thanks @simi1505 — fixed in 179486c, and the diagnosis is worth writing down because the file was never actually missing from the contribution.

`spice/sg13g2_stdcell_hv.spice` is generated, copied by the install script, and is the netlist every one of our characterization and per-cell LVS runs has been using. It was missing from **git**, not from the library: the repo root `.gitignore` carries `*.spice` (line 163), and `libs.ref/.gitignore` re-includes only the `spice/` **directory** — which does not re-include the file inside it. So `git add` skipped it silently in every revision of this branch, with no warning at any point.

Two things kept that hidden:

* `sg13g2_stdcell`'s own spice view predates the `*.spice` rule, so it is already tracked and unaffected — the trap only springs for a library that adds a *new* spice view.
* It reproduces only when cloning the branch. Every test I ran installed onto a PDK tree on disk, where the file is present and LVS passes — which is exactly why my runs were clean while yours died at config load. Good catch; that asymmetry is what made it invisible from my side.

**Fix:** re-include `*.spice` under `libs.ref`, next to the existing directory negations that were clearly meant to have this effect, plus the file itself. Note this touches a shared `.gitignore` rather than only the HV library — @KrzysztofHerman @Mauricio-xx, flagging that as a deliberate repo-level change, since any future library shipping a spice view hits the same trap otherwise. Happy to split it into its own PR if you would prefer that.

No content change to the view: it is the same 84-subcircuit netlist as before, so your verified workaround is no longer needed — and it stays a distinct view from the CDL (`cdl/` is the LVS netlist, `spice/` the simulation netlist used for characterization), matching how `sg13g2_stdcell` ships both.

I also added a check to our install script that asks `git check-ignore --no-index` about every installed file and fails the run if any of them would be invisible to git. It reproduces this exact defect against the unpatched `.gitignore` and passes (191 files) against the fixed one, so this class of silent drop cannot ship again.

Thanks also for the re-test detail — the 0/0 DRC agreement between Magic and KLayout, the native `dfflibmap` mapping and the 2.9× area figure are all useful to have confirmed independently.

One gap still open, which I am working on now: 9 cells have layouts but no liberty entry (the 6 tri-states, the 2 statetable clock gates and `sighold`), exactly as `sg13g2_stdcell` characterizes them. `sighold` is done; the tri-states and clock gates are being measured against ngspice directly, since CharLib has no Hi-Z or statetable input form. I will report separately when that lands.

vcaeken:
Both PDF reports in `libs.ref/sg13g2_stdcell_hv/doc/` are updated to rev. 5 in 7262b34, along with one more characterized cell.

**`sighold` is now characterized.** It was drawn but skipped because CharLib's configurator drops it — its only signal pin is an `inout`, so the "no output pin" branch fires. It also needs no timing characterization at all (the thin-oxide model is `driver_type : bus_hold` plus capacitance and leakage, zero timing groups), so it is measured directly like the tie cells: leakage 0.0118/0.0224 nW, capacitance 0.00367 pF, and a drive-then-release check confirming it holds 3.30000 V and 2.6e-08 V into 1 TΩ.

One methodological note that may interest you, since it contradicts shipped thin-oxide data: charge integration — the library's method for ordinary gate pins — is **invalid** for a bus holder. The integral is dominated by the keeper fighting the driver, so it grows monotonically with the driving edge rate (30.5 fC at a 0.2 ns edge → 86.4 fC at 5 ns, a 2.8× spread). A bias-swept 1 MHz small-signal measurement has no such coupling (7% spread). That also explains `sg13g2_sighold`'s otherwise unexplained 2.8× rise/fall capacitance asymmetry (0.0268 vs 0.0096 pF) as an artifact of its measurement rather than structure — a node's physical capacitance has no direction. The HV cell therefore ships equal rise/fall values and documents the keeper fight-back charge separately as the driver-dependent quantity it is.

**Report updates.** The library report moves to all 84 cells and adds the per-cell layout generators, the strict-rule N-well rebuild, and a new section on LibreLane integration (including the silently-gitignored spice view from the comment above). The characterization-methods report adds a section on the cells no characterizer reaches — separating CharLib's three distinct blind spots, and noting that the tri-state case fails *silently*: the cell is configured, no simulation runs, the empty result is swallowed by `omit_on_failure`, and the log never mentions it — plus the bus-holder result and a section on why absent `max_capacitance`/`max_transition` fails both silently (vacuous STA limit checks) and loudly (the TritonCTS segfault, OpenROAD #11165).

Still open: the 6 tri-states and the 2 statetable clock gates. Both classes are being measured directly against the shipped netlists now — CharLib has no Hi-Z concept and no statetable input form, and lctime carries `three_state` but deliberately pins the enable and skips exactly the enable arcs. Until those tables land, both classes stay excluded from the flow, as the thin-oxide SCL also excludes its own clock gates, so nothing downstream can pick a cell it cannot time.

simi1505:
Re-tested at head 179486c with the same setup as before (8-bit counter macro from the iic-jku SG13G2 AMS chip template, LibreLane Classic flow in IIC-OSIC-TOOLS: LibreLane v3.1.0.dev3, OpenROAD 26Q3-1297, Magic 8.3.589, KLayout 0.30.5, netgen).

**The library now works out of the box — zero local modifications needed.** The previously missing spice view resolved the `CELL_SPICE_MODELS` config failure, and LVS now verifies against the shipped spice models:

- Magic DRC: 0 errors
- KLayout DRC: 0 errors
- Netgen LVS: circuits match uniquely
- STA clean (setup/hold/slew/cap) at 50 MHz, `typ_3p30V_25C`; antenna clean
- Flops map natively via `dfflibmap` (`sg13g2_hv_dfrbpq_1`)

The remaining limitations (single characterized corner, layout-only tristates/clock gates) are documented in the exclude files and didn't affect the flow.

From our side, this is ready to use. Thanks for the fast iterations!

simi1505:
@vcaeken, could you please also build a CMOS5L PR with symlinks now? I will test the other PR as well and merge both PRs into our iic-jku forks.
We will then test your HV Pcells in the August IIC-OSIC-TOOLS release. Maybe there are still some improvements left. For example, additional corners and so on (see follow-up below). 
Then, IHP can decide if they want to have the HV stdcell library officially in their PDK as well. :-)

simi1505:
### sg13g2_stdcell_hv library (this PR / follow-ups)

1. **More liberty corners.** Only `typ_3p30V_25C` exists; fast/slow process and temperature/voltage corners are needed before any multi-corner signoff (the thin-oxide library ships six). Until then, HV timing sign- off is single-corner only.
2. **Switching-power characterization.** No cell has `internal_power` tables, so dynamic power numbers from STA/power reports on HV designs are not meaningful yet (recorded as a limitation in the characterization report).
3. **Characterize the remaining 8 cells**: the 6 tristates (`ebufn_*`/`einvn_*`) and the 2 clock gates (`lgcp_1`/`slgcp_1`) are layout-only and flow-excluded (announced as measured next).
4. **Scan-flop synthesis support.** `sdfbbp_1` stays synth-excluded because Yosys `dfflibmap` cannot pattern-match a scan-mux `next_state`. The shipped `sdfbbp_map.v` covers this opt-in; a short note in the library doc on when/how to enable it (DFT flows) would help users find it.
5. **Guard against gitignored views.** The spice view was silently swallowed by `libs.ref/.gitignore` for five commits. A CI check that every `libs.ref/<lib>/` contains the canonical view set (gds/lef/lib/cdl/spice/verilog) — and that liberty and LEF cell lists match — would catch this class of gap automatically; every one of them produces a confusing downstream failure instead of a clear message.

vcaeken:
**The 6 tri-state cells are now characterized** (bfff921) and enabled in the SCL — `SYNTH_TRISTATE_MAP` and `TRISTATE_CELLS` are set, mirroring the thin-oxide SCL.

These were the *silent* failure in the characterization flow, and worth describing because the mechanism is not obvious: CharLib has no high-impedance concept anywhere in its schema or code, but `gen_charlib_config.py` doesn't skip these cells — `Z` has a `function`, `TE_B` is collected as an ordinary input, and the cell is configured as a two-input gate whose output ignores one input. CharLib emits an empty result, `omit_on_failure` swallows it, and the run log never mentions the cell. lctime is no substitute: it carries `three_state` but deliberately pins the enable and skips exactly the enable arcs. So all three arc classes are measured directly against the shipped netlist.

Validation: 60 tables, 0 empty, 0 monotonicity violations, and the harness is calibrated by re-measuring a shipped cell's pin capacitance through the same code path — `sg13g2_hv_buf_4` pin A lands **0.46%** from the shipped value. Combinational HV/LV delay ratios sit in the library's 2.66× band.

**One deviation you should know about, since it contradicts the thin-oxide tables.** The `three_state_disable` arc cannot be measured the textbook way. Waiting for a released output to cross a voltage threshold is not physically measurable on a capacitively loaded Hi-Z node: once the driver lets go, only the keeper moves the node, so the answer is 0.2·R·C ≈ 0.5 ms and scales with the load. No keeper value fixes it — to make the decay comparable with turn-off it would have to approach the cell's on-resistance, at which point it fights the enabled driver.

The thin-oxide library cannot have measured it that way either: `sg13g2_ebufn_*`'s disable tables are *exactly* load-independent, the seven entries along the load axis differing only by a 1 fs increment — a monotonicity epsilon rather than a measurement. (Its enable tables are similarly clamped at the fast end: `ebufn_4` repeats 0.074755/6/7/8.)

So what is measured here is the quantity that actually goes away — the drive current: time from `TE_B` at 50% to |I_Z| falling through half its on-state value. That is load-independent by construction, needs no synthetic epsilon, and uses the same 50% convention as everything else in the library. Its HV/LV ratio (1.46–1.53) is therefore outside the delay band **by construction**; it is reported rather than tuned, and the tight clustering across two footprints and three drive strengths is what a consistent criterion offset looks like. If IHP has the original disable-arc definition for the thin-oxide library, I would happily match it instead.

Two infrastructure defects were found on the way, both of which would have produced plausible-looking numbers: the mid-rail keeper defining the floating node was left attached during *leakage* measurement, where it sourced its own 1.65 nA into the supply reading — a fictitious 5.48 nW, **150× the true value**; and ngspice needs `OMP_NUM_THREADS=1` here, its spinning OpenMP barriers burning 130+ CPU-seconds without completing against ~4 CPU-seconds single-threaded.

**End-to-end check:** a shared-bus design that infers `$_TBUF_` hardens to 16 `sg13g2_hv_ebufn_2` instances with zero unmapped cells, **Magic DRC 0, KLayout DRC 0, LVS 0**, routing DRC 0, antenna 0, and zero max-slew/max-cap violations against real limits.

Both PDF reports are rebuilt with the derivation, the ratio tables and the two defects above. Remaining: the 2 statetable clock gates, in progress.

simi1505:
### Structure review vs `dev` — convention notes

Compared the PR's file/folder structure at head bfff921 against `dev`'s `sg13g2_stdcell` as the reference. Everything is located correctly (same view subdirs, per-SCL `sg13g2_tech.lef` byte-identical to the thin-oxide copy, `libs.tech` placements match, and the `.gitignore` fix is properly placed). Remaining convention-level observations, none of them misplacements:

1. **Mux maps diverge from `dev`.** The thin-oxide SCL on `dev` no longer ships `mux2_map.v`/`mux4_map.v` (they only exist in older releases), but the HV SCL ships them and sets `SYNTH_MUX_MAP`/`SYNTH_MUX4_MAP`. So on `dev`, thin-oxide synthesis lets ABC map muxes while HV forces `sg13g2_hv_mux2_1`/`mux4_1`. Is that intentional, or should HV follow the newer thin-oxide style? This is now the only structural difference between the two SCL directories apart from `sdfbbp_map.v`.
2. **No `qucs-s/` views.** The thin-oxide library has `sym/qucs-s` and `sch/qucs-s` (plus a `README.md`) alongside `xschem/`; the HV library is xschem-only.
3. **`doc/` PDF naming.** The thin-oxide library ships per-corner datasheets (`sg13g2_stdcell_<corner>.pdf`); the HV library ships a library report and a characterization-methods report instead. Both are useful — a per-corner datasheet (`sg13g2_stdcell_hv_typ_3p30V_25C.pdf`) would additionally match the existing pattern.
4. **`sdfbbp_map.v`** has no thin-oxide counterpart, but the SCL directory is exactly where such a map belongs — fine as is.

simi1505:
### sg13g2_stdcell_hv library (this PR / follow-ups)

1. **More liberty corners.** Only `typ_3p30V_25C` exists; fast/slow process and temperature/voltage corners are needed before any multi-corner signoff (the thin-oxide library ships six). Until then, HV timing signoff is single-corner only.
2. **Switching-power characterization.** No cell has `internal_power` tables, so dynamic power numbers from STA/power reports on HV designs are not meaningful yet (recorded as a limitation in the characterization report).
3. **Characterize the 2 remaining cells**: the statetable clock gates (`lgcp_1`/`slgcp_1`) are still layout-only and flow-excluded — as the thin-oxide SCL also excludes its clock gates, so this may well stay as is.
4. **Scan-flop synthesis support.** `sdfbbp_1` stays synth-excluded because Yosys `dfflibmap` cannot pattern-match a scan-mux `next_state`. The shipped `sdfbbp_map.v` covers this opt-in; a short note in the library doc on when/how to enable it (DFT flows) would help users find it.
5. **Guard against gitignored or silently-empty views.** Two gaps in this PR shared a failure mode — output that never appears and never warns: the spice view swallowed by `libs.ref/.gitignore`, and the tri-state tables silently dropped by CharLib + `omit_on_failure`. A CI check that each `libs.ref/<lib>/` carries the canonical view set, that liberty and LEF cell lists agree, and that no expected cell has zero timing tables would catch this whole class.

vcaeken:
Doc sync in `abaa7be`, plus a rewritten PR description — the prose had fallen behind the artifacts over the last three revisions.

**What was wrong.** The description and `doc/ReleaseNotes.txt` still described the rev. 1 state in places: 68 cells in GDS, 68/68 LVS, "75 cells in the Liberty file", and the 6 tri-states listed as uncharacterized alongside the clock gates. The two PDF reports carried pre-tri-state verification figures (76 cells, 1 932 monotonicity series, 73 output pins with drive limits) and one reference to a companion PDF that is not shipped anywhere.

**What it says now**, re-read from the shipped files rather than from memory:

* GDS + LEF: 84 cells. Per-cell LVS 84/84.
* Liberty: 82 cells, 660 timing tables, 72 of them with arcs; `sighold` as `bus_hold`; 2 tie cells and 7 physical stubs. `verify_lib.py` on the shipped file: 660/660 tables populated, 2 141/2 142 load-axis series monotone (one documented waiver), drive limits on all 79 output pins, 14/14 sequential cells clean.
* The only cells without a Liberty entry are `lgcp_1`/`slgcp_1`, and that is now stated in one place per document rather than three inconsistent ones.
* Numbered cross-references in both reports were pointing at sections that do not exist under `--number-sections`; fixed.

**On your follow-up item 4** (@simi1505): the DFT note is in `doc/ReleaseNotes.txt` §4 and in section 8 of the library report — why `sdfbbp_1` stays synth-excluded (`dfflibmap` cannot pattern-match a scan-mux `next_state`), and the two concrete steps to enable the map: drop it from `synth_exclude.cells`, then set `SYNTH_EXTRA_MAPPING_FILE: pdk_dir::libs.tech/librelane/sg13g2_stdcell_hv/sdfbbp_map.v` at design level, since LibreLane does not let a PDK set that variable.

The single-corner, no-`internal_power` and clock-gate limitations were only in the reports; they are in the release notes now too.

**Structure review items 1 and 3 are open questions for you**, restated in the description: I am happy to drop `mux2_map.v`/`mux4_map.v` and let ABC map muxes to match the newer thin-oxide style on `dev`, and to add a per-corner `sg13g2_stdcell_hv_typ_3p30V_25C.pdf` datasheet alongside the two reports if that naming pattern should be kept. Both are one-line decisions on your side; neither is in this commit.

simi1505:
### Structure review vs `dev` — convention notes (updated)

Compared the PR's file/folder structure at head abaa7be against
`dev`'s `sg13g2_stdcell` as the reference. Everything is located
correctly (same view subdirs, per-SCL `sg13g2_tech.lef` byte-identical
to the thin-oxide copy, `libs.tech` placements match, and the
`.gitignore` fix is properly placed). Remaining convention-level
observations, none of them misplacements:

1. **Mux maps diverge from `dev`.** The thin-oxide SCL on `dev` no longer ships `mux2_map.v`/`mux4_map.v` (they only exist in older releases), but the HV SCL ships them and sets `SYNTH_MUX_MAP`/`SYNTH_MUX4_MAP`. So on `dev`, thin-oxide synthesis lets ABC map muxes while HV forces `sg13g2_hv_mux2_1`/`mux4_1`. Is that intentional, or should HV follow the newer thin-oxide style? This is now the only structural difference between the two SCL directories apart from `sdfbbp_map.v`.
2. **No `qucs-s/` views.** The thin-oxide library has `sym/qucs-s` and `sch/qucs-s` (plus a `README.md`) alongside `xschem/`; the HV library is xschem-only.
3. **`doc/` PDF naming.** The thin-oxide library ships per-corner datasheets (`sg13g2_stdcell_<corner>.pdf`); the HV library ships a library report and a characterization-methods report instead. Both are useful — a per-corner datasheet (`sg13g2_stdcell_hv_typ_3p30V_25C.pdf`) would additionally match the existing pattern.
4. **`sdfbbp_map.v`** has no thin-oxide counterpart, but the SCL directory is exactly where such a map belongs — fine as is.

Resolved since the earlier reviews: `tribuff_map.v` is present and wired (`SYNTH_TRISTATE_MAP`, `TRISTATE_CELLS`), mirroring the thin-oxide SCL; and the shipped `doc/` prose now matches the library. I re-derived the ReleaseNotes figures from the shipped Liberty and they check out: 82 cells, 72 with timing arcs, 79 pins carrying drive limits, and the 10 timing-free cells are exactly the 7 physical stubs plus `sighold` plus the 2 tie cells.

simi1505:
### sg13g2_stdcell_hv library (this PR / follow-ups, updated)

1. **More liberty corners.** Only `typ_3p30V_25C` exists; fast/slow process and temperature/voltage corners are needed before any multi-corner signoff (the thin-oxide library ships six). Until then, HV timing signoff is single-corner only.
2. **Switching-power characterization.** No cell has `internal_power` tables, so dynamic power numbers from STA/power reports on HV designs are not meaningful yet (recorded as a limitation in the release notes and reports).
3. **Characterize the 2 remaining cells**: the statetable clock gates (`lgcp_1`/`slgcp_1`) are still layout-only and flow-excluded — as the thin-oxide SCL also excludes its clock gates, this may well stay as is.
4. **Scan-flop synthesis support.** `sdfbbp_1` stays synth-excluded because Yosys `dfflibmap` cannot pattern-match a scan-mux `next_state`. The shipped `sdfbbp_map.v` covers this opt-in, and the release notes now document how to enable it — worth keeping that note in sync if the exclude list ever changes.
5. **Derive the shipped doc figures from the shipped Liberty.** The last commit had to hand-correct release notes and reports that had fallen two commits behind the library (tri-states and full layout coverage landed before the prose). Generating the cell counts, table counts and limitation lists from the `.lib` at build time would make that drift impossible rather than periodic.
6. **Guard against gitignored or silently-empty views.** Two earlier gaps shared a failure mode — output that never appears and never warns: the spice view swallowed by `libs.ref/.gitignore`, and the tri-state tables silently dropped by CharLib + `omit_on_failure`. A CI check that each `libs.ref/<lib>/` carries the canonical view set, that Liberty and LEF cell lists agree, and that no expected cell has zero timing tables would catch this whole class. (Items 5 and 6 are the same idea applied to prose and to data: assert what you ship against what you built.)

simi1505:
@vcaeken, what do you say? Is this PR ready to be merged into our iic-jku fork? When do you think you'll be finished with the CMOS5L PR? Thanks!

vcaeken:
Good evening, Simon, 

The timing characterization of the remaining two clock-gating cells (sg13g2_hv_lgcp_1 and sg13g2_hv_slgcp_1) is still ongoing, and I would propose that you fork after completion. I am also still working on the PR that adds the SG13CMOS5L symlinks to the library.

I will also generate two more timing corners so that designs can close timing over PVT instead of at the typical corner only, and in particular so that hold violations at the fast/cold corner become visible to the RTL-to-GDS flow:

Setup / max delay: ss, 2.97 V, 125 °C — 204 ps, the slowest of the 27 points, 1.44× the typical corner delay.
Hold / min delay: ff, 3.63 V, −40 °C — 109 ps, the fastest, 0.77× typical.

However, the additional timing corner characterization simulations will likely take a week to complete.

KR KVC

vcaeken:
The PR for SG13CMOS5L has been created: https://github.com/IHP-GmbH/ihp-sg13cmos5l/pull/100

vcaeken:
The clock-gating characterization finished, so the timing side of this PR is complete: **all 84 cells now carry a Liberty entry, 74 of them with timing arcs** (668 delay/transition tables plus the setup/hold constraints).

`sg13g2_hv_lgcp_1` and `sg13g2_hv_slgcp_1` ship the full ICG model rather than a plain latch: `clock_gating_integrated_cell` (`latch_posedge` / `latch_posedge_precontrol`), a `statetable` with the internal state pin, `state_function : "CLK * int_GATE"` on GCLK, the `clock_gate_out_pin` / `clock_gate_clock_pin` / `clock_gate_enable_pin` / `clock_gate_test_pin` attributes, CLK→GCLK propagation, setup/hold of GATE (and SCE) against the rising edge, CLK `min_pulse_width`, and per-state leakage. CharLib has no input form for a statetable, so these two are measured directly against the shipped netlist by `work/char_clockgate/`, the same way the six tri-states and the bus holder were.

One honest note on the data: the `min_pulse_width` of `lgcp_1` and `slgcp_1` is identical point for point, and their CLK→GCLK delays differ only in the 4th–6th digit. That is expected — the scan leg sits on the enable path, not on the clock path — but it does mean the two cells are not distinguishable on the clock arc at this tolerance. It is written up in the release notes rather than smoothed over.

Also in this revision:

* The `spice/` views are now actually tracked. They were being silently dropped by a root `.gitignore` rule (`*.spice`; `!spice/` only re-includes the directory, not the files). Thanks for catching that — the install script now asserts every installed file is trackable, so it cannot regress quietly.
* The reports and release notes are synced to the library state.

Timing corners are running now: **fast (3.60 V, −40 °C, `mos_ff`)** and **slow (3.00 V, 125 °C, `mos_ss`)**, both with full 84-cell coverage, to go alongside the existing typical corner. The flow is corner-parameterized (`work/corners.py`, `work/run_corner.sh`) and the PDK-level LibreLane config registers whichever corners are present, so they drop in without further structural change. I will push them as they complete — that is a large simulation job, so please do not block the fork on it if you would rather take the library at one corner now and the other two as a follow-up.

vcaeken:
**The fast timing corner is in: `sg13g2_stdcell_hv_fast_3p60V_m40C.lib`** (3.60 V, −40 °C, `mos_ff`), all 84 cells, 668 timing tables, on the same slew and load grids as the typical corner so STA can interpolate across them. The PDK-level LibreLane config registers both corners and pairs each with the matching IO liberty (`sg13g2_io_fast_1p65V_3p6V_m40C` for the fast one); `nom_typ_3p30V_25C` stays the default.

The **slow corner (3.00 V, 125 °C) is not yet characterized** and is running now. The release notes say so plainly rather than implying full PVT coverage — anything that must close over all corners should treat the library as incomplete until it lands. Everything else is unchanged, so this is safe to fork at whatever point suits you.

Two things worth reporting honestly, since both concern data quality rather than plumbing.

**Building a second corner exposed four latent defects**, every one invisible while typ was the only corner that had ever existed — because typ was built up incrementally over many sessions, never once from an empty directory. The most serious: the tri-state characterizer's on-disk measurement cache was keyed by cell/slew/direction with **no corner in the key**, so the fast run silently replayed typ's simulations. It was caught only by an independent physical check — a keeper-hold assertion noticing a floating node sitting at 3.3 V when the rail was 3.6 V. The others were a calibration reference frozen at typ (which reported a broken harness that was in fact correct to 0.43 %), a step that patched Liberty entries nothing in the flow ever created, and two non-idempotent patchers that would double-insert on the documented resume path. All fixed, and the reports now document the failure modes for anyone adding a third corner.

**The clock-gate measurements had to be redone.** 17 of 98 CLK→GCLK delay grid points and 7 of 16 min-pulse-width bisections were destroyed by simulation timeouts, and the pinned bisections would have told STA these clock gates need a ~48 ns pulse — every realistic clock reported as too narrow. The guards refused to emit, which is what they are for. The cause turned out to be our own thread-to-core ratio: ngspice runs ~8 threads per deck on this machine despite `num_threads=1` in `.spiceinit` and `OMP_NUM_THREADS=1` in the environment. Re-run with fewer concurrent jobs, every point measured cleanly, and every re-measured `min_pulse_width` came out *smaller* than its typ counterpart — the physically correct direction for a faster corner. Why the thread pinning is ignored is not yet understood; it is recorded as a follow-up rather than papered over.

`verify_lib` passes on the fast corner with **no waiver at all** — the `xnor2_1` load-axis non-monotonicity that typ carries as a documented, pessimistic outlier simply does not occur at this corner.

vcaeken:
**All three PVT corners are now characterized**, so the library covers full PVT:

| corner | file | models |
|---|---|---|
| typ | `sg13g2_stdcell_hv_typ_3p30V_25C.lib` | `mos_tt` |
| fast | `sg13g2_stdcell_hv_fast_3p60V_m40C.lib` | `mos_ff` |
| slow | `sg13g2_stdcell_hv_slow_3p00V_125C.lib` | `mos_ss` |

Each covers all 84 cells with 668 timing tables, on identical slew and load grids so STA can interpolate across them, and `verify_lib` passes on all three. The LibreLane config registers all three and pairs each with the matching `sg13g2_io` corner; `nom_typ_3p30V_25C` stays the default. The 3.6 / 3.0 V rails follow the PDK's own 3.3 V domain rather than a strict ±10 %, so the stdcells agree with the pad ring they sit next to.

**One correction to data already on this branch, which I would rather flag than bury.** The antenna-diode leakage deck left its signal pin dangling with no DC path, so the operating point was under-determined. At 125 °C ngspice refused it outright — gmin stepping, source stepping and the transient op all failed — and that failure is what exposed it. The quiet damage was at the corners where it *did* converge:

```
antennanp cell_leakage_power (nW)
  typ    0.21561    -> 0.0325726      (6.6x too high, already pushed)
  fast   0.0295047  -> 0.0387641
  slow   (no convergence) -> 0.0270255
```

Every signal pin in that deck now gets a 1 GΩ tie to ground (3 nA, three orders below the currents being measured). The corrected values order with supply voltage; the floating ones did not. Only `antennanp` was affected — every other physical-only cell has just VDD and VSS. It is a small number on a `dont_use` cell, so no timing is affected, but it was wrong and it shipped.

**Two process notes**, since they say something about how much of this flow was actually exercised:

* The HV/LV ratio cross-check had **never run**. Its parser required `") ;"` where the thin-oxide library writes `");"`, so it matched nothing and raised. Because it sits in the clock-gate driver's task list, its crash killed `run_corner.sh` under `set -e` before the merge and the final two stages — which means **no corner had ever completed a scripted end-to-end run**; I had finished the tail by hand each time without noticing. Both are fixed.
* Running the slow corner with a lower job count from the start avoided the simulator oversubscription that cost the fast corner two rounds of rework: delay 98/98 with no holes, min-pulse-width 16/16 with none pinned, setup/hold 18/18, all first pass.

Now that the cross-check works it reports, for the clock gates, that HV `min_pulse_width` is roughly flat across input slew (0.46 → 0.97 ns) while the thin-oxide equivalent climbs steeply, giving HV/LV ratios from 0.24 to 5.47. I have not chased that down — the HV values are monotonic in slew and order correctly across corners, so it reads as a difference in how the two libraries define the constraint rather than an error, but you should know it is there.

vcaeken:
**Qucs-S views added**, bringing this library to view parity with `sg13g2_stdcell`: `sch/qucs-s` (84 schematics), `sym/qucs-s` (84 component XML plus the 45 shared gate-shape `.sym`), and the `sch/` and `sym/` READMEs.

They are a device-level retarget of the thin-oxide schematics — identical topology and wiring, devices changed to `sg13_hv_*` and resized **from the shipped SPICE netlist rather than by a ratio**, because the thick-oxide transform is not a single scale factor: gate length goes 130 nm → 450 nm in the common case, but 180 nm is clamped to 450 nm, 250 nm becomes 625 nm, and the deliberately long devices (500 nm, 700 nm, 1 µm) are left alone. `sg13g2_hv_tiehi` and `sg13g2_hv_tielo` are composed from this library's own `inv_1` instead, since the thick-oxide tie cells are a tied-input inverter rather than the thin-oxide four-transistor chain.

`work/verify_qucs.py` gates them three ways: the 82 retargeted cells are diffed against their thin-oxide originals and may differ only in device size and title-block fields (a changed wire or coordinate fails, which proves the geometry was never touched); all 920 device sizes are checked against the SPICE netlist; and the two composed tie cells get a real netlist extraction from the wire graph, compared terminal by terminal.

To expose them as Qucs-S components, the symbol directory needs the same kind of link the thin-oxide library uses:

```sh
ln -sfn ../../libs.ref/sg13g2_stdcell_hv/sym/qucs-s \
        "$PDK_ROOT/$PDK/libs.tech/qucs-s/symbols_stdcell_hv"
```

I have deliberately **not** added that symlink in this PR — whether the PDK ships per-library links, or leaves it to `install.py`, is a maintainer decision, and a loop over `libs.ref/*/sym/qucs-s` in `install.py` would make it self-maintaining as libraries are added. Happy to add either form if you prefer.

---

**Two environment notes for anyone testing this branch**, neither caused by it:

1. **Qucs-S standard cells may not resolve in older iic-osic-tools images.** In the image I am using, `libs.tech/qucs-s/symbols_stdcell` still points at `/home/rahman/temp/IHP-Open-PDK/...` and dangles, so *no* standard cells — thin- or thick-oxide — appear as Qucs-S components. That is simply a PDK snapshot (27 Jul) predating #1097, which made the link relative on `dev` on 11 Aug; nothing to fix here, but worth knowing if the views appear to be missing when you try them.

2. **ngspice ignores `OMP_NUM_THREADS`.** It reads its own `num_threads` from `.spiceinit` and calls `omp_set_num_threads()` itself, overriding the environment. The container ships `~/.spiceinit` with `num_threads=4`, so a job-parallel characterization run silently asks for jobs × 4 threads. That is what corrupted 17 of 98 clock-gate delay points and 7 of 16 min-pulse-width bisections in my first pass at the fast corner — a timed-out trial inside a bisection reads as "failed", so the search returns its own bound as though measured. Only relevant if you re-run the characterization, but it is invisible when it happens, which is why I mention it.

Because Qucs-S is unavailable in this image (binary off `PATH`, and the symlink above), **I have not been able to confirm these schematics open and netlist in the Qucs-S GUI.** They are structurally correct and gated against the netlist, but that is a weaker claim, and worth a look before merge.

simi1505:
Tested at head 40f1f02. The two new corner files do not parse, so any
multi-corner HV run stops at STA:

```
Error: .../sg13g2_stdcell_hv_slow_3p00V_125C.lib line 3893 function syntax error.
```

Cause: on the sequential cells, the fast and slow files write the `function` attribute as an assignment instead of a boolean expression referencing the state variable.

fast and slow:
```
    pin (Q) {
      function : "Q = D" ;
```

typ, which is correct and parses:
```
    pin (Q) {
      function : "IQ" ;
```

Occurrences of the assignment form: typ 0, fast 21, slow 21. It affects the 14 flip-flops and latches (dfrbp*, dfrbpq*, sdfrbp*, sdfbbp_1, dlhq_1, dlhr*, dllr*). Four variants appear, `Q = D`, `Q_N = !(D)`,
`Q = ((SCE & SCD) | (!(SCE) & D))` and its inverse.

To confirm the diagnosis I replaced only those 21 lines per file with `function : "IQ"` or `function : "IQN"` and reran. That alone is enough. Our 8-bit counter then hardens clean on all three corners:

| corner | worst hold slack (ns) | violations |
|---|---|---|
| nom_typ_3p30V_25C | 0.7708 | 0 |
| nom_fast_3p60V_m40C | 0.5536 | 0 |
| nom_slow_3p00V_125C | 1.1870 | 0 |

Magic DRC 0, KLayout DRC 0, LVS matches uniquely, setup and hold clean.

One thing to look at while fixing this: the new files declare two `ff` groups per cell, `ff (IQ, IQinv)` and `ff (IQN, IQNinv)`, where typ declares a single `ff (IQ,IQN)`. It parses after the fix, but it points at the same generator difference.

Thanks for adding the corners. Once this is in, hold signoff on the HV library becomes possible for the first time.

vcaeken:
Confirmed and fixed — thank you, that was an unusually actionable report. The line number, the per-file count and the four variants meant the diagnosis took minutes rather than an afternoon.

**Root cause.** Both new corners shipped CharLib's raw sequential emission. `work/fix_lib_seq.py` exists precisely to repair it and had simply never been run over them: the combinational path applies it automatically inside `run_charlib.sh`, but the sequential stage writes `seq_<corner>.lib` and merges that directly, so it bypassed the repair. The typ corner was fixed the same way months ago — which is why typ shows 0 and the two new files show 21 each.

You are right that the double `ff` groups are the same generator difference, and the same pass removes them. Per file, before → after:

| | before | after |
|---|---|---|
| assignment-form `function` | 21 | **0** |
| `ff` groups | 14 | **9** |
| `latch` groups | 7 | **5** |
| cells | 84 | 84 |

So 21 pin functions and 7 redundant state groups per corner. **No numeric data is touched at all** — zero timing, capacitance or leakage lines differ between the broken and repaired files; it is purely the state-declaration syntax.

Your instinct about it being more than cosmetic is right too: a flip-flop whose `Q` is a function of `D` is combinational, which is exactly what made `dfflibmap` fail before typ was repaired.

Fixed on the branch. All three corners now read in OpenSTA with zero errors and pass `verify_lib`.

**On why this got past us, which is the part that bothers me more than the bug.** `verify_lib.py` reads the Liberty as text by design — structure, cross-view, areas, monotonicity, drive limits, sequential arcs, special-cell constructs. Every one of those passed on a file that no timer can parse. The checker and the artifact shared an assumption, so the check could not see the defect.

`verify_lib.py` now runs `read_liberty` in OpenSTA as its first check, and I verified it in both directions: it FAILs on the as-committed file at line 3893 and passes on the repaired one. That is the gate that should have existed from the start, and it would also have caught this at the fast corner a day earlier.

Sorry for the round trip — you should not have had to find this by running it.

simi1505:
Retested at head 62f4c27. The fix works, and the library now hardens a design cleanly with no local changes of any kind.

Our 8-bit counter macro ran the full LibreLane Classic flow on all three corners with 50MHz:

| corner | worst hold (ns) | worst setup (ns) | violations |
|---|---|---|---|
| nom_typ_3p30V_25C | 0.7708 | 14.6098 | 0 |
| nom_fast_3p60V_m40C | 0.5536 | 14.9593 | 0 |
| nom_slow_3p00V_125C | 1.1870 | 13.9537 | 0 |

Magic DRC 0, KLayout DRC 0, netgen LVS matches uniquely, antenna clean, and all three per-corner power reports are produced.

Worth noting beyond the pass or fail: the corners order physically. Worst hold lands at the fast corner and tightest setup at the slow corner, which is the check that separates a real corner set from one that merely parses. This is the first HV run where hold signoff actually means something.

Rendered layout of the hardened counter:

counter_top_hv_white

The one limitation we still see is the absence of `internal_power` tables, so power reports from an HV run are not yet meaningful. Everything else we reported over the past few days is now resolved.

Thanks for the quick turnaround on the corner fix.

simi1505:
### Structure review vs `dev` — convention notes (updated for 62f4c27)

Rechecked the file and folder structure against `dev`'s `sg13g2_stdcell`. Everything is correctly located: the same view subdirs, per-SCL `sg13g2_tech.lef` byte-identical to the thin-oxide copy, and correct `libs.tech` placements. Two convention-level observations remain, neither of them a misplacement:

1. **Mux maps diverge from `dev`.** The thin-oxide SCL on `dev` ships `config.tcl`, `latch_map.v`, `tribuff_map.v`, the two exclude lists and `tracks.info`. It has no `mux2_map.v` or `mux4_map.v`, but the HV SCL ships both and sets `SYNTH_MUX_MAP` and `SYNTH_MUX4_MAP`. So thin-oxide synthesis lets ABC map muxes while HV forces `sg13g2_hv_mux2_1` and `sg13g2_hv_mux4_1`. Is that intentional? It is the only functional difference left between the two SCL directories, apart from `sdfbbp_map.v`.
2. **No per-corner datasheet.** The thin-oxide library ships one PDF per corner (`sg13g2_stdcell_<corner>.pdf`). The HV library ships a library report and a characterization-methods report instead, now in Markdown as well as PDF, which is more useful content. A per-corner datasheet would additionally match the existing pattern, and there are three corners to describe now.

`sdfbbp_map.v` has no thin-oxide counterpart, but the SCL directory is exactly where such a map belongs, and the release notes explain when to enable it. That one is fine as is.

Resolved since the earlier reviews: `tribuff_map.v` is present and wired, the qucs-s `sym/` and `sch/` views landed in 6b473d2, and the shipped doc now matches the library.

simi1505:
### Future improvements (updated for 62f4c27)

1. **Switching-power characterization.** No cell carries `internal_power` tables in any of the three corners, so power reports from an HV run are not meaningful. This is the last functional gap we can see.
2. More ss, ff, and tt timing corners at 1.2V and 1.5V.
3. **Derive the shipped doc figures from the shipped Liberty.** Two separate commits have now had to hand-correct release notes and reports that had fallen behind the library (abaa7be and 673dc96). Generating cell counts, table counts and the limitation list from the `.lib` files at build time would end that class of drift rather than repeating the correction each time the library grows.
4. **Keep extending the parser-level gate.** Having `verify_lib.py` ask a timer to read each Liberty was the right fix for the corner regression, because no text-level check could have caught it. Two neighbours worth the same treatment, since both have already bitten this PR: assert that every `libs.ref/<lib>/` carries the canonical view set (the spice view was silently gitignored for five commits), and that no expected cell has zero timing tables (the tri-states were silently dropped by CharLib and hidden by `omit_on_failure`). A useful extra assertion for a multi-corner library is that the corners order physically, worst hold at fast and tightest setup at slow.

### Resolved since the last list

Multi-corner Liberty (357a5ad, ac0c261), clock-gate characterization so all 84 cells are timed (41487ef), the corner sequential-emission repair (62f4c27), the qucs-s views (6b473d2), and documenting how to enable `sdfbbp_map.v` for DFT flows.

simi1505:
Thanks, @vcaeken. For now, this looks good to me. I will merge it into our iic-jku fork. It will be available in the August IIC-OSIC-TOOLS release.

vcaeken:
Pushed `c72a676`: the verilog view now ships `sg13g2_udp.v` as a symlink.

@b10346 pointed out on #1207 that the HV cellsets carry no UDP file. He is
right, and it applies here too: `sg13g2_stdcell_hv.v` instantiates six `ihp_*`
user-defined primitives (`ihp_dff_r`, `ihp_dff_sr_1`, `ihp_latch`,
`ihp_latch_r`, `ihp_mux2`, `ihp_mux4`) and defines none of them, so nothing
with a flop or a latch elaborated standalone.

```
libs.ref/sg13g2_stdcell_hv/verilog/sg13g2_udp.v
  -> ../../sg13g2_stdcell/verilog/sg13g2_udp.v
```

A symlink rather than a copy: the `ihp_*` names are shared across the SG13G2
libraries, so a second definition collides the moment a flow reads both, and a
copy can drift from the original. One file does neither, and a flow still
includes exactly one `*_udp.v`. `sg13cmos5l_stdcell` already ships its own
`sg13cmos5l_udp.v` beside its cells, so this also matches the tree's existing
layout.

Verified: `iverilog -g2012 sg13g2_udp.v sg13g2_stdcell_hv.v` elaborates with 0
errors and 0 unresolved modules, from the `verilog/` directory alone.

The release note, the cell report and the worked example all described the
absence of the file as a deliberate choice; that was wrong and is corrected in
the same commit.

Separately, while reproducing his other finding: `yosys read_verilog` cannot
parse these views — but nor can it parse IHP's own `sg13g2_stdcell.v`, at the
same constructs and in the same numbers (68 `ifnone`, 17 `primitive`). It is a
Yosys frontend gap, not something this PR introduces, and it does not affect
synthesis, which maps against the Liberty. Filed as #1224.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1108, opened 2026-08-18
cap_cmomi: via the tooth tips and re-bill the model to the cell

Mauricio-xx:
This PR adds the missing tip via to each tooth of the cap_cmomi PCell and corrects the Verilog-A model to bill what the cell actually draws. The same correction lands in the sg13cmos5l twin.

Three model corrections follow. The coupled-row count was billed as floor(w/0.89)-1; the -1 excluded end fingers that this bar-terminated cell does not draw, so every device was one coupled row short, about 1% at large widths and up to a factor of two at the minimum width. The single-side (feed=same) term is re-derived from a per-N table to a pad-length fit. And the opposite-side (double) feed, previously assumed to cancel and dropped to zero, is now billed explicitly at 0.152*pad_len; on the default feed=double device that is about +2.9% (size-dependent, roughly +8% at the minimum width down to +0.1% at the largest).

The area-density coefficients do not change: they were fit to a fully-viaed cell (teeth viaed at base and tip), so adding the tip vias brings the drawn geometry up to what the density law already assumed rather than changing it. What the missing vias had masked was the feed residual, now billed.

The updated model was cross-checked against a Palace EM campaign; the structures and results are in a separate documentation repository: https://github.com/Mauricio-xx/CAP_MOMI-EM-/tree/main

Geometry and consistency tests pass; the regenerated LVS testcase layouts differ from the previous ones only by the added tip vias.

Following image shows the changes in terms of Capacitance vs area 
model_grid_l_w


Mauricio-xx:
Hi @sergeiandreyev , do you have any changes request related with the pycell of this new device? old pcell is actually merged, but this pr includes some changes on the physical structure of the device and if you have any observation related with pycell code I can do the amends before merge and port this to the cmos5l companion pr (github.com/IHP-GmbH/ihp-sg13cmos5l/pull/98)

Makefile
sergeiandreyev: very weird comment text, please change the wording

Mauricio-xx:
@simi1505 @hpretl FYI

sergeiandreyev:
ok, some comments on the Pcell:
- label:
  - should be placed inside cell boundary (ref. `cmim`, `rfcmim`, etc.)
  - instead of 'model name' use 'device name'
  - use two lines for the name and cap value
- cap value should be added as a Pcell parameter, not only as a label (ref. `cmim`, `rfcmim`, etc.)
- do not see any `techparams` in Pcell code.. global technology constants should be added to JSON tech file, and taken from there if using defaults inside a Pcell (i.e. no `VIA_CUT = 0.19`, instead `VIA_CUT = techparams['V1_a']`, etc.)
- was the cell origin position reviewed by someone?
- sorry, I've missed when there was a discussion about layers: why we had `MIM.*` and now we want to have `Recog.mom*`? 
- maybe a check enabled would be better for the last parameter? any chance of minimize that long line?
image

Mauricio-xx:
Thanks @sergeiandreyev, all done on the g2 PCell, and mirrored in the cmos5l twin ([ihp-sg13cmos5l#98](https://github.com/IHP-GmbH/ihp-sg13cmos5l/pull/98)):

- Label: two lines centred inside the cell, the device name + C value (as in cmim/rfcmim).
- C value: added as a PCell parameter, computed from the model.
- techparams: the constants that are existing DRC rules now come from the tech file, matching cmim/via_stack: via cut = `V1_a`, tooth via pitch = `V1_a + V1_b`, tooth extension = `UC_Y - T_BAR/2 - Mn_b`, pin width = `Mn_a`. The MoM-specific ones (unit-cell pitch, brick stagger, feed pad sizes) are design choices that satisfy rules rather than rules themselves, and no pcell in the tree stores that kind of device-specific layout in the tech JSON, so they stay as commented constants. Values are unchanged, so the testcase layouts are byte-identical (XOR empty on every layer).
- Origin: now centred on the device centre. Tell me if that's the convention you meant.
- Layers: this is a metal MoM on Metal1..Metal5, never on MIM (that's the MIM-cap dielectric cmim/rfcmim use). The marker is Recog.mom (99/39), which LVS keys on, so nothing regressed.
- subblock: now a boolean checkbox, and the long feed description is shortened.

LVS and the consistency check pass with the regenerated layouts.

ihp-sg13g2/libs.tech/klayout/tech/sg13g2.lyp
```diff
@@ -3136,8 +3136,8 @@
     <source>41/36</source>
   </properties>
   <properties>
-    <frame-color>#ffffff</frame-color>
-    <fill-color>#ffffff</fill-color>
+    <frame-color>#000000</frame-color>
```
sergeiandreyev: `Text.drawing` colors changed unexpectedly

sergeiandreyev:
image
is this long line expected?

hpretl:
> image is this long line expected?

Hopefully not :-) Looks like the Tcl evaluation fails.

simi1505:
Hi @Mauricio-xx, I wanted to kindly ask when you plan to finish and merge this PR? We want to release the August IIC-OSIC-TOOLS image, but we really want to have a working and validated MOM cap in G2 and CMOS5L PDK in this release. Do you think it will be finished by tomorrow? Thanks!

KrzysztofHerman (approved):

sergeiandreyev:
@KrzysztofHerman, @Mauricio-xx, in my view this PR is still not ready to be merged, unless you have already started on follow-up PRs w/ fixes - please review my feedback

PR #1115, opened 2026-08-24
DNM: Introduce REUSE

dnltz:
REUSE is an FSFE specification that makes copyright and licensing information machine-readable by requiring every file to declare its copyright holder and SPDX license identifier. Files that can carry comments get an SPDX header inline; binaries and other files that cannot are annotated centrally in REUSE.toml, and the full text of each referenced license is stored in LICENSES/. The `reuse lint` tool then verifies that no file is left without this information, which is what the new license-check workflow runs on every pull request.

-----

This is just an example how the SPDX headers look like. If IHP think it's worth using REUSE, I would migrate all other files to SPDX headers files and add annotations for binaries (like GDS).

Why it's worth using REUSE?
- Automatic checks if files have a wrong/missing license
- Covers binaries files
- Allows tracking of licenses used in large chip design projects

-----

The CI action will now fail as example how it would look like if files have missing license headers.

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -1,0 +1,6 @@
+version = 1
+
+[[annotations]]
+path = "ihp-sg13g2/libs.ref/*/gds/*.gds"
+SPDX-FileCopyrightText = "2024 IHP PDK Authors"
+SPDX-License-Identifier = "Apache 2.0"
--- a/.github/workflows/license-check.yaml
+++ b/.github/workflows/license-check.yaml
@@ -1,0 +1,35 @@
+# SPDX-FileCopyrightText: 2024 IHP PDK Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# This workflow checks for missing license or copyright identifiers
+
+name: PDK License Check
+
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+    branches:
+      - 'dev'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+
+      - name: Install REUSE
+        run: |
+          python -m pip install --upgrade pip
+          pip install reuse
+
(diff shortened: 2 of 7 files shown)


dnltz:
@sergeiandreyev example how license management could be simplified. It also removes these legacy license headers in all files :)

KrzysztofHerman:
@dnltz I will present it on Friday to introduce it. How does it work for binaries ?

REUSE.toml
```diff
@@ -0,0 +1,6 @@
+version = 1
+
+[[annotations]]
+path = "ihp-sg13g2/libs.ref/*/gds/*.gds"
```
dnltz, line 4: @KrzysztofHerman I added GDS files here as example. You basically maintain multiple annotations here for different sections.

sergeiandreyev:
quite extensive report :)
and already found an issue - in some GNUCAP models we have AGPL-3.0, @KrzysztofHerman:
image
which needs to be considered carefully since it's a strict copyleft license which overrides our Apache 2.0 (there is plenty of information on this on the Internet)

still some comments:
- for GitHub actions we should limit the run only to the files included in the PR, not running on full PDK
- is it possible to exclude binaries, `.md`, etc. where it doesn't make sense (or am I missing smth?)

I'll look into it in more details myself.. thank you for sharing this project!

dnltz:
> quite extensive report :) and already found an issue - in some GNUCAP models we have AGPL-3.0, @KrzysztofHerman: image which needs to be considered carefully since it's a strict copyleft license which overrides our Apache 2.0 (there is plenty of information on this on the Internet)
> 
> still some comments:
> 
>     * for GitHub actions we should limit the run only to the files included in the PR, not running on full PDK
I assume that could be done, yes.
> 
>     * is it possible to exclude binaries, `.md`, etc. where it doesn't make sense (or am I missing smth?)
Yes, I left a comment because @KrzysztofHerman asked the same question. You can cover files in the REULSE.toml file and wilcards are also possible (in case you want to assign a license to a whole directory).
> 
> 
> I'll look into it in more details myself.. thank you for sharing this project!

PR #1120, opened 2026-08-24
 Fix crash in sprintf utility function

Ayoubbelguellaoui:

- [x] Tests pass

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/ihp/utility_functions.py
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/ihp/utility_functions.py
@@ -18,7 +18,6 @@
 
 import math
 from math import sqrt
-from io import StringIO
 
 from cni.dlo import Tech, Numeric
 from cni.dlo import Orientation, Location, Layer
@@ -170,9 +169,7 @@ def zerop(value):
 # sprintf
 #***********************************************************************************************************************
 def sprintf(fmt, *args):
-    buf = StringIO.StringIO()
-    buf.write(fmt % args)
-    return buf.getvalue()
+    return fmt % args
 
 #***********************************************************************************************************************
 # strcat
--- a/ihp-sg13g2/libs.tech/klayout/sg13g2_tests/sprintf_test.py
+++ b/ihp-sg13g2/libs.tech/klayout/sg13g2_tests/sprintf_test.py
@@ -1,0 +1,33 @@
+########################################################################
+#
+# Copyright 2026 IHP PDK Authors
+#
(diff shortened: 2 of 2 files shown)


sergeiandreyev:
Hi @Ayoubbelguellaoui, how did you run into this crash? could you please share a testcase? where do you use it?
I cannot find anywhere in the OpenPDK where this function is used.. ~~if so we should just delete it :)~~
UPD: ok, it could be used in future for Transmission Line (TL) Pcells

Ayoubbelguellaoui:
> Hi @Ayoubbelguellaoui, how did you run into this crash? could you please share a testcase? where do you use it?
> I cannot find anywhere in the OpenPDK where this function is used.. ~~if so we should just delete it :)~~
> UPD: ok, it could be used in future for Transmission Line (TL) Pcells
> 

I found the crash by inspecting sprintf(). It imports StringIO with from io import StringIO, then calls StringIO.StringIO(), which raises AttributeError.
I confirmed this by calling sprintf() directly.
No callers of sprintf() exist yet in OpenPDK.
I'll add a regression test that exercises the function directly...

sergeiandreyev:
@Ayoubbelguellaoui, would you like to include regression in this PR? otherwise I'm ok to merge it

sergeiandreyev (approved):
made the regression w/ Codex

Ayoubbelguellaoui:
I've just added the regression test in the latest commit.

sergeiandreyev:
> I've just added the regression test in the latest commit.

I've added `pcell4klayout` api to the Python path, otherwise I had an error:
```
$ python3 sprintf_test.py 
Traceback (most recent call last):
  File "/home/andreev/opensource/IHP-Open-PDK.test/ihp-sg13g2/libs.tech/klayout/sg13g2_tests/sprintf_test.py", line 27, in <module>
    from sg13g2_pycell_lib.ihp.utility_functions import sprintf
  File "/home/andreev/opensource/IHP-Open-PDK.test/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py", line 23, in <module>
    from cni.tech import Tech
ModuleNotFoundError: No module named 'cni'
```

sergeiandreyev:
Hi @Ayoubbelguellaoui, are you ok with the above updates? Can you reproduce the issue on my side?
if yes, I will merge the PR

Ayoubbelguellaoui:
Yes, I confirmed and reproduced the issue on my side. Without adding api_directory to sys.path, running sprintf_test.py via plain python3 fails with ModuleNotFoundError: No module named 'cni'.

With your update (8b517619), the path resolves correctly and the regression test passes seamlessly (sprintf regression test passed).

The changes look great to me—feel free to proceed with merging the PR  
Thanks for adding this fix!

PR #660, opened 2025-08-27
`lvsres` device

sergeiandreyev:
Current status:
- [x] Pycell

PR #1121, opened 2026-08-25
`lvsres` device (restored from #660)

### LINKED ISSUE #1041: [PCells] Implementation of a metal resistor `Rmetal` (for LVS of RFICs)
As far as I know, there is no metal resistor PCell available. This makes LVS of purely metal structures, for example, RF structures, impossible. Hence the question: Are there internal discussions about whether a metal resistor PCell will be added to the IHP Open-PDK?

My proposal for adding such a metal resistor would be:
- Make a symbol / netlist for Xschem and ngspice for a metal resistor. We could reuse existing resistors and adapt the settings for it. We mainly need W, L and R. If we just use this resistor for LVS, then R could be fixed (e.g. 1uOhm), and only W and L are changeable. But in general, a fully working metal resistor would be desirable.
- Make a PCell in KLayout. This is actually a simplified version of the via PCells. We need a drop-down menu to select the metal layer and three input fields for L, W and R. In the end, just a rectangle with W x L is drawn on the selected layer. Pretty simple, I would say. Again, if we just use this resistor for LVS, then R could be fixed (e.g. 1uOhm) and only W and L are changeable. But in general, a fully working metal resistor would be desirable. In the end, both R in the schematic and layout have to be the same for a successful LVS. But when there is a fixed formula for this, I do not think that this is an issue.
- In the end, this PCell has to land in LVS, DRC, PEX for KLayout and Magic. This is probably the most critical part since different maintainers are involved here. But I hope together we can figure this out.

For us, this PCell would be highly needed, since we are currently doing a lot with passive RF structures and without it, no full top-level LVS of our open-source RFICs is possible. Thanks!

Of course, we can support points 1 and 2. For LVS, DRC, and PEX, it is probably more complicated.

FYI, @hpretl, @martinjankoehler, @p-fath, @RTimothyEdwards, @KrzysztofHerman, @d-m-bailey, @VolkerMuehlhaus, @Mauricio-xx, @yavarhadi

sergeiandreyev:
Continuing #660: latest agreement is that `metRes` will be a separate device, not a replacement

Current status:
- [x] KLayout Pycell
- [ ] KLayout LVS update (`res_<metal_type>` -> `lvsres ... layer=<metal_type>`)
- [ ] Xschem symbol
- [ ] Qucs-S symbol
- [ ] Documentation update

Fixes #1041

### DIFF
--- a/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
+++ b/ihp-sg13g2/libs.tech/klayout/python/sg13g2_pycell_lib/__init__.py
@@ -76,7 +76,8 @@ moduleNames = [
         'moscap_p_code',
         'pnpMPA_code',
         'isolbox_code',
-        'schottky_code'
+        'schottky_code',
+        'lvsres_code'
 ]
 
 def getProcessNames():
@@ -281,4 +282,3 @@ class SG13G2_NativePCellLib(pya.Library):
 # instantiate and register the libraries
 PyCellLib()
 SG13G2_NativePCellLib()
-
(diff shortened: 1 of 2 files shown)


sergeiandreyev:
@simi1505, FYI

simi1505:
Thanks, @sergeiandreyev!
@hpretl, @p-fath, @schwarzj99, FYI.

Issue #1125, opened 2026-08-27
LVS regressions are broken


dnltz:
**Describe the bug**
While migrating the CMOS5L PDK over, I figured out the LVS regression tests are broken.

I would like to propose to ALWAYS run LVS and DRC regressions tests and remove the filter inside the GitHub workflow. 

**Why?**

Regression can also fail if files changed which are not part of the regression context (e.g. new KLayout version, CDL files, ...). However, regressions will only fail next time someone changes the tests itself.

PR #1126, opened 2026-08-27
ihp-sg13g2: libs.tech: klayout: Fix LVS testing

dnltz:
LVS regression testing currently fails. Fix cap_cmomf test to pass again.


dnltz:
@KrzysztofHerman @martinjankoehler I need your feedback on this change. LVS Regression Testing is currently failing and Claude generated this fix.

I also opened #1125 and think we should lways run regression testing.

PR #1127, opened 2026-08-28
libs.tech: klayout: tech: scripts: sealring: Make height available

dnltz:
The parameter "height" should still be available and set the new variable "length". Thus, existing scripts don't break and the height->length change is not a breaking change in upcoming releases.
    
Fixes c4cd27c8 ("libs.tech: klayout: tech: scripts: sealring: Rename height to length")


dnltz:
hey @sergeiandreyev, @mole99 just reported to me the fallback is missing. This PR adds `height` back as fallback for `length` - as requested by you and somehow disappeared from the original PR :o

sergeiandreyev (approved):

PR #1136, opened 2026-09-01
SRAM shared location

dnltz:
Move all IHP SRAM macros to ihp-common/libs.ref, which servers as common location for all PDKs. Add symlinks to the SG13G2 and SG13CMOS5L PDKs.

This change doesn't introduce any breaking changing and the same SRAM macros are still available in both PDKs.


dnltz:
@KrzysztofHerman is there anything blocking this PR?

Issue #805, opened 2026-01-30
Flicker noise duplicated in my integrated noise report


DiarmuidSLICE:
Hello,

Attached shows the total integrated noise for a hv_nmos in IHP. As you can see, its flicker noise appears duplicated. 

<img width="1406" height="88" alt="Image" src="https://github.com/user-attachments/assets/b6a74bae-317d-407e-85c4-fa68c97767b5" />

Why is this?

Thanks,

Diarmuid

[tb_nmos_integrated_noise.txt](https://github.com/user-attachments/files/24956250/tb_nmos_integrated_noise.txt)

dwarning:
There are two different flicker noise contribution in the psp model - see here from the model:

            I(DI,SI)     <+  flicker_noise(MULT_i * Sfl, EF_i, "flicker");
            I(DI,SI)     <+  flicker_noise(MULT_i * Sfledge, EFEDGE_i, "flicker");

Unfortunately they have same name "flicker". 
This is correct and not an issue. You can close please.

DiarmuidSLICE:
Ah that explains it. Very interesting. And the contributions are so similiar. So total flicker is a root sum of these then. 

Clears things up for me. 

Thanks for clarifying. 

Closing ticket now.

PR #1138, opened 2026-09-02
psp103: give the edge part of flicker noise a unique name

dwarning:
A unique name was given for the edge part of the ids flicker noise contribution. 

This small PR will solve the problem of #805. There is still a bug in ngspice osdi noise interface in summing up the different noise contributions which will be solved in next release 48.

### DIFF
--- a/ihp-sg13g2/libs.tech/verilog-a/psp103/Changelog
+++ b/ihp-sg13g2/libs.tech/verilog-a/psp103/Changelog
@@ -1,3 +1,7 @@
+Change name of flicker noise contribution edge transistor from "flicker" to "flickeredge"
+to distinguish between bottom and edge part. psp103_module.include line 1954
+This is inline with User’s Manual- NXP / CEA-Leti v. 103.8.2
+
 Xyce/ADMS adaptions:
 Use IHP specific code for AD, AS, PD, PS calculation.
 Remove $mfactor from OP output variables calculation
--- a/ihp-sg13g2/libs.tech/verilog-a/psp103/PSP103_module.include
+++ b/ihp-sg13g2/libs.tech/verilog-a/psp103/PSP103_module.include
@@ -1951,7 +1951,7 @@ analog begin
             I(GP,DI)     <+  white_noise(MULT_i * shot_igd, "igd");
             I(BS,SI)     <+  white_noise(MULT_i * jnoise_s, "ibs");
             I(BD,DI)     <+  white_noise(MULT_i * jnoise_d, "ibd");
-            I(DI,SI)     <+  flicker_noise(sigVds * MULT_i * Sfledge, EFEDGE_i, "flicker");
+            I(DI,SI)     <+  flicker_noise(sigVds * MULT_i * Sfledge, EFEDGE_i, "flickeredge");
             I(DI,SI)     <+  white_noise(MULT_i * sqidedge * sqidedge, "ididedge");
 
         end // noise


felix-salfelder:
Dear Dietmar. Please make these/such changes conditional to the intended use case.

    `ifndef __NGSPICE__ // or similar
       // normal noise here
    `else
      // hacked noise here
    `endif

This way it won't interfere with existing workflows and the patch will highlight and explain the aberration.

Thanks

dwarning:
Hallo Felix,
this is not a ngspice specific problem. The model authors gave the two noise sources over same branch same name. And you can see in mentioned issue #805 what is happen in the report. This will also happen in gnucap or any other simulator.
It is just a inaccuracy.

felix-salfelder:
On Wed, Sep 02, 2026 at 01:46:45AM -0700, dwarning wrote:
> this is not a ngspice specific problem. The model authors gave the two
> noise sources over same branch same name. And you can see in mentioned
> issue #805 what is happen in the report. This will also happen in
> gnucap or any other simulator.  It is just a inaccuracy.

I do not understand. "This is not an ngspice [..] problem" conflicts
with "summing up the different noise contributions which will be solved
in next release 48" you wrote earlier. Issue 805 is about ngspice
afaics.

Gnucap implements noise following the standard. The standard says the following
(in Section "4.6.4.2 flicker_noise" and others).

"""
The optional name argument acts as a label for the noise source used
when the simulator outputs the individual contribution of each noise
source to the total output noise. The contributions of noise sources
with the same name from the same instance of a module are combined in
the noise contribution summary.
"""

I believe the module is correct the way it is without your patch.

Thanks.

dwarning:
I think best way to understand the problem is, that you simulate the attached test case with gnucap and the unpatched model and publish here the printout.
[nmos_noise.txt](https://github.com/user-attachments/files/31734414/nmos_noise.txt)

KrzysztofHerman:
@dwarning please add to the PSP model at least an annotation what was changed. Ideally change also the revision number. 
I will try to reach out to the CEA-Leti developers to report this change.

felix-salfelder:
To understand the problem.. a standard compliant simulator will compute different values for "flicker" noise after applying the proposed patch. This can be bad, if looking at "flicker" noise is part of your workflow.

This is irrespective of what a manual says, and does not (really) depend on a simulator choice. Hence, to avoid problems, counting up the model revision number is a good approach.

Model changes should ideally be caught in regression tests. Imagine you had a regression test... We have regression tests for psp103 and noise, but missing out on a "flicker" test. This should be completed when folding the noise tests into the PDK. @deutzlukas-hub ?

dwarning:
I don't like this kind of theoretical discussion. If you follow my proposal and simulate my testbench with gnucap then you would understand the issue. It seems you also not looked into #805.
To naming conventions see latest User’s Manual- NXP / CEA-Leti version 103.8.2, p. 5.

felix-salfelder:
Dear Dietmar.

You should not not make assumptions about what I understand. If you are (partly) right, please make the effort to dig up a valid justification for your patch.

I do not need your testbench. I know how noise analysis works in Gnucap, because I implemented it.

It was you who claimed that #805 reports a problem with ngspice. Looking into it, I found that you were right. Noise contributions are not added as they should have been. Beyond that, I could not read your mind.

If your patch is a pre-emptive bugfix from a future psp103 (which is information you withheld) then it may be best to wait for a new revision. Upstream authors might also explain to you the theory behind revision numbers and the need for consistency.

KrzysztofHerman (approved):

PR #1156, opened 2026-09-02
Move KLayout scripts to ihp-common

### LINKED ISSUE #1152: sealring X-Axis  Y-Axis inconsistency in sealring_code.py and searling.py
Hello,

We have the x and y inconsistency problem in sealring creation. Right now, if you give 
sealring.py -rd width=1000 -rd height=2000, it ends up generating the sealring with x=2000 and y=1000
the reason is in sealring_code.py:
        specs('l', defL, 'Length(X-Axis)')
        specs('w', defW, 'Width(Y-Axis)')

but in searling.py :
    :param width: Width (X-Axis) of the sealring.
    :param height: Heigth (Y-Axis) of the sealring.
...
    pcell = layout.add_pcell_variant(lib, pcell_decl.id(), {'w': f'{width}u', 'l': f'{heigth}u'})
...

so the width got swapped from x to y when creating the pcell in sealring.py.

Suggestion:
If possible, please don't use Width/Length/Heigth as variables, let's be consistent and use x_axis, y_axis in all the scripts, so there will be no confusing naming. 
Thanks a lot.
Zerun Jiang

dnltz:
Move all three KLayout scripts to ihp-common, add symlinks to each PDK and use a thiny wrapper to call them.

This change doesn't break any existing user. However, it changes two things:
- ihp-sg13cmos5l's sealring.py flips X/Y axis to match the PCell.
- ihp-sg13g2's bondbad.py now has the option to generate LEF.

Fixes #1152


dnltz:
@mole99 - shared scripts between both PDKs. This also introduces all recent sealring.py changes made in G2 to the CMOS5L PDK.

mole99:
@dnltz How about a different approach that doesn't rely on wrapper scripts:

We can determine the current active tech within KLayout, which is either `sg13g2` or `sg13cmos5l`. Based on this, the scripts in `ihp-common` adjust the max metal layer, etc.

This would allow us to simply create symlinks from the PDK variants to `ihp-common`, without the need for wrapper scripts.

dnltz:
> @dnltz How about a different approach that doesn't rely on wrapper scripts:
> 
> We can determine the current active tech within KLayout, which is either `sg13g2` or `sg13cmos5l`. Based on this, the scripts in `ihp-common` adjust the max metal layer, etc.
> 
> This would allow us to simply create symlinks from the PDK variants to `ihp-common`, without the need for wrapper scripts.

@mole99 can we because @simi1505 added the tech name in this PR: https://github.com/IHP-GmbH/IHP-Open-PDK/pull/943

I can try again and ideally we can get the currently used TECH.

mole99 (approved):
Looks good to me. It prevents duplication of the files, and LibreLane's full chip flow still works as before.

dnltz:
@KrzysztofHerman PR #1215 was not required because this PR makes it obsolete again. Please review this PR :)

dnltz:
@KrzysztofHerman is there anything blocking this PR? The `length` parameter is missing in the CIEL package right now :(

Issue #100, opened 2024-05-02
RPPD extraction (alpha LVS)


bmurmann:
A series combination of RPPD (as shown below) matches the schematic only with option --no_simplify. But this option won't work for many other parts of a real circuit. Any remedies that can be implemented?

My test cases has 26 resistors in series in the layout and 26 resistors in series in the schematic. When any simplifications are turned on, there is no more match, since the derived simplifications between sch and lay differ.

This could be a general KLayout issue?

![image](https://github.com/IHP-GmbH/IHP-Open-PDK/assets/62622760/2af8068f-7a8d-427a-b75f-29abc7af3d1a)

sergeiandreyev:
Hello @bmurmann, could you please confirm that with the latest LVS update the issue is fixed?

bmurmann:
Confirmed, this is fixed.

PR #1207, opened 2026-09-02
libs.ref: add sg13cmos5l_stdcell_hv (links to the G2 thick-oxide cells)

vcaeken:
Re-homed from IHP-GmbH/ihp-sg13cmos5l#100 at @KrzysztofHerman's request — that repository is being archived and its content now lives here under `ihp-sg13cmos5l/`. Same change, same author, new home.

## What

`sg13cmos5l_stdcell_hv` exposes the thick-oxide (3.3 V) standard cells added by #1103 under CMOS5L names, so a CMOS5L design can synthesise 3.3 V logic without reaching into the SG13G2 tree by hand.

**It contains no copied data.** 19 relative symlinks into `ihp-sg13g2/libs.ref/sg13g2_stdcell_hv`, so the two PDKs cannot drift apart — the same technique this repository already uses for `ihp-sg13cmos5l/libs.ref/sg13cmos5l_sram`.

| path | contents |
|---|---|
| `libs.ref/sg13cmos5l_stdcell_hv` | 5 view links, 3 Liberty links (typ / fast / slow), 3 directory links, the shared tech LEF |
| `libs.tech/librelane/sg13cmos5l_stdcell_hv` | complete SCL — 7 links plus its own `config.tcl`, declaring all three corners each paired with the matching `sg13cmos5l_io` corner, and a `tracks.info` derived from `sg13cmos5l_tech.lef` (M1–M4 X 0.48 / Y 0.42, TopMetal1 1.64/3.28) |
| `libs.tech/klayout/tech/pymacros/sg13cmos5l_stdcell_hv.lym` | cell-library registration, PDK-relative GDS path |
| `libs.tech/xschem/xschemrc` | four lines, alongside the existing CMOS5L entries |
| `libs.ref/.gitignore` | see below |

## Why the move simplifies it

The generator (`make_cmos5l_pr.py` in the provenance repo) always assumed *a checkout root containing both PDK trees*. That is exactly this repository's layout, so **the relative link targets are unchanged from #100** — the migration cost nothing but a different checkout to run against.

## `libs.ref/.gitignore`

Added for the same reason `ihp-sg13g2` needed one in #1103: the root `.gitignore`'s `*.spice` rule silently drops a newly added spice view, and `!spice/` alone re-includes only the directory, not the files in it. Without it, `sg13cmos5l_stdcell_hv/spice/` never reaches the PR and nothing warns you.

I checked that it changes the visibility of **nothing that already exists** — `sg13cmos5l_stdcell`'s own spice view predates the rule and is already tracked, which is precisely why this only surfaces for a library adding a new one.

## Verification

In a `dev` checkout with #1103 applied:

* **19/19** symlinks resolve inside the PDK root
* **21/21** configured paths exist
* OpenSTA reads **all three** Liberty files through the links with **zero errors**

## Dependency — why this is a draft

This depends on #1103, which is what adds `ihp-sg13g2/libs.ref/sg13g2_stdcell_hv`. Until that merges into `dev`, every link here dangles. I have opened it as a draft and will mark it ready as soon as #1103 lands; happy to keep it that way for as long as suits your review order.

One open question inherited from #100: I link **per file**, so the names read `sg13cmos5l_stdcell_hv_*`, consistent with how `sg13cmos5l_stdcell` names its files. The `sg13cmos5l_sram` precedent instead symlinks the whole directory, which is simpler but keeps the `sg13g2_*` filenames and would need the LibreLane `LIB` patterns adjusted. Happy to switch if you prefer the directory form.

Provenance: generated and verified by the scripts at https://github.com/ChipDesign-BV/sg13g2_stdcell_hv (Apache-2.0).


vcaeken:
All three findings were valid and are fixed in `0c13920`. Thank you — the second one in particular was a genuine internal contradiction I had walked past several times.

**The important part: all three were also present in #1103.** This port derives its KLayout macro, corner block and xschemrc lines from the SG13G2 library, so a review of one was effectively a review of both. I fixed them in the generators rather than in the generated files, and pushed the corresponding corrections to #1103 as `6de5e6a`.

**1. `DEFAULT_CORNER` outside `STA_CORNERS`** — correct, and self-contradictory as written. The `foreach` above it guards every corner with `file exists` precisely so a partial install degrades gracefully, and then the default was set to typ unconditionally. A checkout carrying only fast and slow would name a corner that is not registered. Now:

```tcl
if { [lsearch -exact $_hv_corners "nom_typ_3p30V_25C"] >= 0 } {
    set ::env(DEFAULT_CORNER) "nom_typ_3p30V_25C"
} elseif { [llength $_hv_corners] > 0 } {
    set ::env(DEFAULT_CORNER) [lindex $_hv_corners 0]
}
```

**2. `SG13CMOS5L_HV_HOME` semantics** — correct, and worse in this library than in the G2 one, exactly as you noted: the code joined `$HOME/gds/` while the message said "repository root", and here the GDS sits under `libs.ref/sg13cmos5l_stdcell_hv/gds/`, so following the instruction could never find it. The override now accepts **either** the library directory **or** a PDK root containing `libs.ref/`, and the failure message names both instead of guessing for the reader.

**3. Unguarded xschemrc paths** — correct, and inconsistent with the file it patches: that `xschemrc` already guards `PDK_ROOT` itself with `file isdir`. The point is sharper here than in the G2 library, since these paths are symlinks into `ihp-sg13g2` and do not resolve at all until #1103 lands. Both additions are now wrapped:

```tcl
if { [file isdir ${PDK_ROOT}/${PDK}/libs.ref/sg13cmos5l_stdcell_hv/sym/xschem] } {
  append XSCHEM_LIBRARY_PATH :${PDK_ROOT}/${PDK}/libs.ref/sg13cmos5l_stdcell_hv/sym/xschem
  set ::SG13G2_HV_SCH ${PDK_ROOT}/${PDK}/libs.ref/sg13cmos5l_stdcell_hv/sch/xschem
}
```

so xschem starts clean both before and after the dependency lands.

**Verified after the change**: all four touched Tcl files are complete (`info complete`), 19/19 symlinks resolve, 21/21 configured paths exist, and on the #1103 side 84/84 cells still netlist through the installed symbols with all 410 files trackable.

This stays a draft until #1103 merges into `dev`.

b10346:
@vcaeken I have a question regarding the intended use of the 3.3 V standard-cell library in a complete digital flow.

As far as I understand, there is currently no I/O library providing a 3.3 V I/O ↔ 3.3 V core interface, and no level shifters between the low-voltage core domain and the 3.3 V domain.

How is this library intended to be used in a complete chip-level digital flow? In particular, how should the 3.3 V logic interface with the pad ring or with the existing low-voltage core logic?

ihp-sg13cmos5l/libs.ref/sg13cmos5l_stdcell_hv/verilog/sg13cmos5l_stdcell_hv.v
b10346, line 1: yosys> read_verilog ~/github/IHP-Open-PDK-PR1207/ihp-sg13cmos5l/libs.ref/sg13cmos5l_stdcell_hv/verilog/sg13cmos5l_stdcell_hv.v 

1. Executing Verilog-2005 frontend: /home/balashov/github/IHP-Open-PDK-PR1207/ihp-sg13cmos5l/libs.ref/sg13cmos5l_stdcell_hv/verilog/sg13cmos5l_stdcell_hv.v
Parsing Verilog input from `/home/balashov/github/IHP-Open-PDK-PR1207/ihp-sg13cmos5l/libs.ref/sg13cmos5l_stdcell_hv/verilog/sg13cmos5l_stdcell_hv.v' to AST representation.
/home/balashov/github/IHP-Open-PDK-PR1207/ihp-sg13cmos5l/libs.ref/sg13cmos5l_stdcell_hv/verilog/sg13cmos5l_stdcell_hv.v:43: ERROR: syntax error, unexpected '=', expecting TOK_IGNORED_SPECIFY_AND or ',' or ')' or '['

b10346, line 1: verilog model and links on user defined primitives is not presented in g2_hv and cmos5l_hv cellcets. Please see 1.2V library folder verilog.

ihp-sg13cmos5l/libs.tech/librelane/sg13cmos5l_stdcell_hv/tracks.info
b10346, line 1: Looks wrong. M1 - horizontal routing direction, vertical pitch is 0.42. X=0.48 Y=0.42

b10346, line 1: TopMetal1 2.28 is not correct. Correct is 3.28

b10346 (commented):
I have checked links of CDL, GDS, Verilog, LEF, Liberty.
Content of these files. Some issues were found. Please check comments regarding LEF, track info and verilog models.
It can not be approved at this moment.

vcaeken:
> @vcaeken I have a question regarding the intended use of the 3.3 V standard-cell library in a complete digital flow.
> 
> As far as I understand, there is currently no I/O library providing a 3.3 V I/O ↔ 3.3 V core interface, and no level shifters between the low-voltage core domain and the 3.3 V domain.
> 
> How is this library intended to be used in a complete chip-level digital flow? In particular, how should the 3.3 V logic interface with the pad ring or with the existing low-voltage core logic?

Good evening, Alexey,

The I/O cells and level shifters need to be designed.

There are two use cases for the library:
- The cells can be used in analog mixed-signal blocks (buffers, encoders, decoders, etc.).
- The library can be used for RTL2GDS of a ULP digital block. At the same supply voltage, the thick-oxide (I/O) logic gates have less static leakage than the thin-oxide (core) logic gates.

I will correct the issues, which you have reported.

With kind regards, KVC

vcaeken:
Thanks @b10346 -- all four are real and all four are fixed in `506d8e9`.

### `tracks.info` (X 0.48 / Y 0.42, TopMetal1 3.28)

You are right, and the reason is worth writing down because it was not what I
assumed. The file was not mine: it was a verbatim copy of
`libs.tech/librelane/sg13cmos5l_stdcell/tracks.info`, taken on the reasoning
that the CMOS5L grid should come from the CMOS5L library.

That file disagrees with the tech LEF shipped beside it:

| | `sg13cmos5l_stdcell/tracks.info` | `sg13cmos5l_tech.lef` |
|---|---|---|
| Metal1 | `X 0.0 0.42` / `Y 0.0 0.42` | `PITCH 0.48 0.42` |
| Metal2 | `X 0.0 0.48` / `Y 0.0 0.48` | `PITCH 0.48 0.42` |
| Metal3 | `X 0.0 0.42` / `Y 0.0 0.42` | `PITCH 0.48 0.42` |
| Metal4 | `X 0.0 0.48` / `Y 0.0 0.48` | `PITCH 0.48 0.42` |
| TopMetal1 | `1.64 2.28` | `PITCH 3.28 3.28`, `OFFSET 1.64 1.64` |

A 0.42 µm Metal1 X pitch also does not divide the 0.48 µm `CoreSite` the cells
sit on. **This looks like a bug in the thin-oxide library itself** -- I have
not touched it, since it is outside this PR, but it seems worth a look, and if
you would like a separate issue for it I will open one.

The generator now **derives** `tracks.info` from `sg13cmos5l_tech.lef` rather
than copying from either neighbour. LEF gives `DIRECTION`, `PITCH x y` and
`OFFSET x y` per routing layer, which is exactly a `tracks.info` row pair, and
it is the file the router obeys. Result, which is your values:

```
Metal1 X 0.0 0.48     Metal1 Y 0.0 0.42
Metal2 X 0.0 0.48     Metal2 Y 0.0 0.42
Metal3 X 0.0 0.48     Metal3 Y 0.0 0.42
Metal4 X 0.0 0.48     Metal4 Y 0.0 0.42
TopMetal1 X 1.64 3.28 TopMetal1 Y 1.64 3.28
```

Copying G2's file instead would have produced the right pitches -- these are
the G2 cells and the two LEFs declare identical ones -- but it also declares
Metal5 and TopMetal2, which the M1–M4–TM1 stack does not have. Deriving from
the LEF needs no such luck, and it is cross-checked numerically against the G2
grid on the five layers both stacks share, so a LEF parsed wrongly cannot pass
quietly.

### The UDP view

Also right, and this one was a straightforward omission. The cells instantiate
six `ihp_*` user-defined primitives (`ihp_dff_r`, `ihp_dff_sr_1`, `ihp_latch`,
`ihp_latch_r`, `ihp_mux2`, `ihp_mux4`) and defined none of them, so nothing
with a flop or a latch elaborated standalone.

Both libraries now ship the UDP file as a **symlink**, not a copy:

```
#1103  libs.ref/sg13g2_stdcell_hv/verilog/sg13g2_udp.v
         -> ../../sg13g2_stdcell/verilog/sg13g2_udp.v
#1207  libs.ref/sg13cmos5l_stdcell_hv/verilog/sg13cmos5l_udp.v
         -> ../../sg13cmos5l_stdcell/verilog/sg13cmos5l_udp.v
```

A copy would put a second definition of the shared `ihp_*` names in the tree,
which collides the moment a flow reads both libraries, and could drift. One
file can do neither. `iverilog -g2012 <lib>_udp.v <lib>_stdcell_hv.v` now
elaborates with 0 errors and 0 unresolved modules, from the `verilog/`
directory alone. The release note that presented the missing file as a
deliberate choice was wrong and is corrected in both PRs.

### The Yosys parse error

Reproduced, and it is a Yosys frontend limitation rather than something either
PR introduces -- IHP's own thin-oxide library fails at the same constructs:

```
$ yosys -p "read_verilog libs.ref/sg13g2_stdcell/verilog/sg13g2_stdcell.v"
.../sg13g2_stdcell.v:165: ERROR: syntax error, unexpected '('     # if (…) inside specify
$ yosys -p "read_verilog -specify …/sg13g2_stdcell.v"
.../sg13g2_stdcell.v:42:  ERROR: … unexpected TOK_IFNONE          # 68 of these
$ yosys -p "read_verilog …/sg13g2_udp.v"
.../sg13g2_udp.v:19:      ERROR: syntax error, unexpected TOK_ID  # primitive ihp_latch (…)
```

Identical counts in the HV views (68 `ifnone`, 17 `primitive`), and Icarus
reads all of it with 0 errors. It does not block synthesis, which maps against
the Liberty, but it does rule the Verilog view out of anything Yosys-based.
Filed as #1224 with the options I can see, including generating a
Yosys-readable second view if you would prefer that.

### One question

Your summary mentions the LEF, but I could not find an inline comment on a LEF
file -- if one did not post, could you repeat it? I would rather fix it in this
round than after.

`#1207` stays a draft until `#1103` merges into `dev`; every link into
`ihp-sg13g2/libs.ref/sg13g2_stdcell_hv` dangles until then.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1206, opened 2026-09-02
Automate synchronization checks for forked CMOS5L files


KrzysztofHerman:
**Describe the bug**
The current CI pin-bump checks only whether symlinks resolve. It does not inspect the 294 forked, non-symlinked CMOS5L files that also have G2 counterparts. Changes to those files can therefore accumulate silently between the two repositories.

**Suggested Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Critical

**Environment**
Git; CMOS5L-to-G2 forked-file mapping; `.github/ihp-sg13g2.ref`; CI; G2 history; 294 forked files identified by the issue #15 audit.

**To Reproduce**
1. Change `.github/ihp-sg13g2.ref` to a newer G2 revision.
2. Review the resulting CI checks.
3. Inspect forked CMOS5L files with G2 counterparts for commits between the old and new pins.
4. Observe that CI reports symlink validity but does not identify relevant changes in forked files.

Source issue: https://github.com/iic-jku/ihp-sg13cmos5l/issues/15

Issue #1205, opened 2026-09-02
Add the magic bondpad LEF regression test


KrzysztofHerman:
**Describe the bug**
G2 contains `libs.tech/magic/test_bondpad_lef.sh`, while CMOS5L does not. Bondpad LEF generation and import can therefore regress without detection.

**Suggested Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Critical

**Environment**
magic; CMOS5L bondpad PCell and LEF flow; `libs.tech/magic/test_bondpad_lef.sh`; CI and Makefile test entry points.

**To Reproduce**
1. Inspect `libs.tech/magic/` in the CMOS5L repository.
2. Search for `test_bondpad_lef.sh`.
3. Generate or import a bondpad LEF.
4. Observe that no repository regression test checks the generated geometry and pins.

Source issue: https://github.com/iic-jku/ihp-sg13cmos5l/issues/15

Issue #1204, opened 2026-09-02
Add root Makefile help, lint, gnucap, and PCell-sweep targets


KrzysztofHerman:
**Describe the bug**
CMOS5L lacks the root Makefile targets available in G2 for `help`, `lint`, `test-gnucap`, and PCell sweeps. Developers do not have consistent discovery or convenient entry points for these checks.

**Suggested Priority**
- [x] Low
- [ ] Medium
- [ ] High
- [ ] Critical

**Environment**
CMOS5L root Makefile; lint tools; gnucap tests; KLayout PCell sweep infrastructure; G2 Makefile.

**To Reproduce**
1. Run `make help` in the CMOS5L repository.
2. Attempt to run the `lint`, `test-gnucap`, and PCell-sweep workflows through root Makefile targets.
3. Observe that the targets are absent or require undocumented manual commands.

Source issue: https://github.com/iic-jku/ihp-sg13cmos5l/issues/15

Issue #1203, opened 2026-09-02
Add an RTL-to-GDS CI workflow


KrzysztofHerman:
**Describe the bug**
The CMOS5L repository has no `rtl-to-gds.yml` workflow. LibreLane configuration, technology data, and standard-cell changes can therefore break the complete RTL-to-GDS path without automated detection.

**Suggested Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Critical

**Environment**
GitHub Actions; LibreLane; CMOS5L standard-cell libraries and technology files; RTL-to-GDS flow.

**To Reproduce**
1. Inspect `.github/workflows/` in the CMOS5L repository.
2. Search for `rtl-to-gds.yml` or an equivalent flow workflow.
3. Submit a change that breaks synthesis or routing configuration.
4. Observe that CI has no end-to-end RTL-to-GDS job to detect the regression.

Source issue: https://github.com/iic-jku/ihp-sg13cmos5l/issues/15

PR #1209, opened 2026-09-03
CI: run each PDK's checks only when that PDK is affected

### LINKED ISSUE #1158: Update the CMOS5L G2 reference pin
**Describe the bug**
The CMOS5L CI pin in `.github/ihp-sg13g2.ref` points to `d2cc0355`, which is behind the G2 revision used for the audit. Fixes present in G2 therefore cannot reach the CMOS5L build or regression jobs. The pin should be updated through a reviewable synchronization change that accounts for commits affecting forked CMOS5L files.

**Suggested Priority**
- [ ] Low
- [x] Medium
- [ ] High
- [ ] Critical

**Environment**
G2 reference `d2cc0355`; CMOS5L commit `548d9c1`; CI workflows and symlinked files in the CMOS5L repository.

**To Reproduce**
1. Check out the CMOS5L repository at `548d9c1`.
2. Read `.github/ihp-sg13g2.ref`.
3. Compare the pinned revision with the current intended G2 tip.
4. Observe that newer G2 changes are not available to CMOS5L CI or inherited files.

Source issue: https://github.com/iic-jku/ihp-sg13cmos5l/issues/15

Mauricio-xx:
The repo now holds two PDKs, ihp-sg13g2 and ihp-sg13cmos5l. cmos5l reuses g2's files through symlinks, so a g2 change can affect cmos5l, but a cmos5l change never affects g2. Today the workflows run both PDKs on any change, and changes to the shared build files (pdk.mk, Makefile, versions.txt) trigger nothing. This PR fixes both.

What runs now:
- a change under ihp-sg13g2 runs g2 and cmos5l
- a change to a cmos5l-only file runs cmos5l
- a change to a shared file runs every PDK

How it works: one map file, `.github/pdk-ci.yml`, lists each PDK and what it depends on. A small, unit-tested script turns the changed files into the list of PDKs to run, and the DRC, LVS and lint workflows read that list. Each workflow ends in one always-run "gate" job. Adding a third PDK is one entry in the map.

Also new: the KLayout PCell tests now run in CI (library build, concurrent load, PCell-DRC, native PCells, sprintf), and a separate manual/weekly workflow runs the heavy suites (cap sweeps, cap_cmomi model, gnucap).

Two old test bugs were fixed along the way: the PCell DRC and native tests called a removed API, and the cap sweeps did not accept the maximum metal-density rules.

Branch protection (needs a repo admin, after merge): mark the per-workflow "gate" jobs as the required checks (DRC, LVS, lint, PCell), not the per-PDK jobs under them. Each "gate" job always runs and reports a single pass or fail for its whole workflow, so it still reports when a change skips a PDK. A per-PDK job does not run when its PDK is not affected, and a required check that never runs would block every PR.

Closes #1158, #1204.

Provides the CI base that these issues build on: #1203 (RTL-to-GDS workflow), #1189 (qucs-s integration), #1205 (magic bondpad LEF test), and part of #1171, #1182, #1183, #1185 and #1206.

Follow-ups surfaced while building this CI (tracked separately, not fixed here):
- #1210: the sealring PCell fails to coerce its default length "6.0u".
- #1211: the PCell tests can report success even when the run failed; they become real gates once #1210 is fixed and the tests fail on ERROR.
- #1212: the cmos5l DRC regression counts rules that were never tested as a pass; harden together with #1184.

### DIFF
--- a/.github/pdk-ci.yml
+++ b/.github/pdk-ci.yml
@@ -1,0 +1,59 @@
+# ==========================================================================
+# Copyright 2026 IHP PDK Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# SPDX-License-Identifier: Apache-2.0
+# ==========================================================================
+
+# Single source of truth for which PDK CI runs on a given change.
+#
+# This repository holds more than one PDK in one tree. They are not independent:
+# ihp-sg13cmos5l reuses large parts of ihp-sg13g2 through relative symlinks (its
+# DRC/LVS rule decks, testcases, pycell library, models, ...), so a change to
+# ihp-sg13g2 can break ihp-sg13cmos5l, while nothing in ihp-sg13g2 depends on
+# ihp-sg13cmos5l. The dependency is one-directional.
+#
+# .github/scripts/affected_pdks.py reads this file and turns a set of changed
+# files into the list of PDKs whose CI must run:
+#   - a change under a PDK's own directory affects that PDK, plus every PDK that
+#     depends on it (the reverse of `depends_on`);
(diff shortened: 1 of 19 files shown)


dnltz:
Seriously, don't do that!

Run all checks every time something changes in this PDK. I already reported https://github.com/IHP-GmbH/IHP-Open-PDK/issues/1125 because this workflow was only running when changing the LVS regression itself.

If you're afraid you're running into the GitHub action limit, get a server at IHP and run all checks as self-hosted in a DMZ.

This PR is introducing a monster of dependencies no one can track. Simplify it by always running checks.

